PR #6, opened 2014-12-29
reference updates

PBarmby:
Entry with CHECKME shoudl be checked (it's the same as the one in hg-novice).


jhamrick:
This looks good to me too :+1:

PR #7, opened 2015-01-05
Add requirements.txt necessary to build and run validator

tomwright01:
Copied requirements.txt from lesson-template repository


gvwilson:
Good catch - I've folded this into lesson-template/core, and will merge from there.

PR #15, opened 2015-01-17
 Small fix to conflict lesson

r-gaia-cs:
Students don't need to change directory now that they are working with a partner.


gvwilson:
Please separate the change to the `.md` file from the update to the other `.html` files.

r-gaia-cs:
I update the others HTML files in #17. This pull request should be OK now.

Issue #29, opened 2015-01-21
Issues at 01-backup.md


r-gaia-cs:
From https://raw.githubusercontent.com/swcarpentry/lesson-template/gh-pages/tools/check.py:

- [x] Header section is missing expected row 'minutes'
- [x] Heading at line 28 should be level 2
- [x] Heading at line 75 should be level 2
- [x] Heading at line 128 should be level 2
- [x] Heading at line 273 should be level 2
- [x] Heading at line 510 should be level 2
- [x] Heading at line 595 should be level 2
- [x] Heading at line 692 should be level 2
- [ ] Document contains heading not specified in the template: Setting Up
- [ ] Document contains heading not specified in the template: Creating a Repository
- [ ] Document contains heading not specified in the template: Tracking Changes to Files
- [ ] Document contains heading not specified in the template: Changing a File
- [ ] Document contains heading not specified in the template: Exploring History
- [ ] Document contains heading not specified in the template: Recovering Old Versions
- [ ] Document contains heading not specified in the template: Ignoring Things
- [ ] The topic page should not have sub-headings outside of special blocks. If a topic needs sub-headings, it should be broken into multiple topics.

iglpdc:
Partly fixed in #37. I marked as completed the corresponding tasks above.

iglpdc:
I want to fix the issues with this file to conform with the lesson template. The problem is that to do that either this topic is broken into several parts (I see three: setup and git init, add-diff-commit cycle, and exploring history), or we simply remove the headers separating the different subtopics (I think this will make a very long, unstructured file). 

Thoughts?

gvwilson:
Hi Ivan,
Please break it into pieces - it really wants to be smaller chunks. 
That'll mean renumbering subsequent files...

iglpdc:
Good, I will.

Issue #28, opened 2015-01-21
Issues at 08-open.md


r-gaia-cs:
From https://raw.githubusercontent.com/swcarpentry/lesson-template/gh-pages/tools/check.py:

- [x] Header section is missing expected row 'minutes'
- [ ] Heading at line 160 should be level 2
- [ ] Heading at line 267 should be level 2
- [ ] Document contains heading not specified in the template: Version control as electronic lab notebook
- [ ] Document contains heading not specified in the template: Licensing
- [ ] Document contains heading not specified in the template: Licensing for non-software products
- [ ] Document contains heading not specified in the template: Hosting
- [ ] Document contains heading not specified in the template: Can I Use an Open License?
- [ ] The topic page should not have sub-headings outside of special blocks. If a topic needs sub-headings, it should be broken into multiple topics.

iglpdc:
I've updated @r-gaia-cs 's note on top of this discussion to reflect the current status.

Issue #27, opened 2015-01-21
Issues at 02-collab.md


r-gaia-cs:
From https://raw.githubusercontent.com/swcarpentry/lesson-template/gh-pages/tools/check.py:

- [ ] Header section is missing expected row 'minutes'

iglpdc:
Closing this, now aligns to lesson template by #38, but opening a new issue, as we don't have an estimation for the lesson duration.

Issue #26, opened 2015-01-21
Issues at 03-conflict.md


r-gaia-cs:
From https://raw.githubusercontent.com/swcarpentry/lesson-template/gh-pages/tools/check.py:

- [x] Header section is missing expected row 'minutes'

iglpdc:
Closing this, now aligns to lesson template, but opening a new issue, as we don't have an estimation for the lesson duration.

PR #38, opened 2015-02-09
Align with lesson & markdown template spec

abought:
Address parts of #26, #27, #28, #29. Adds `minutes` fields to headers (with placeholder "FIXME" value pending input by others)

Also converts  and <a> tags to markdown (so that styles automatically get added by the build process/template where appropriate).

Did not rebuild the HTML, as that would have included additional changes beyond those contained in this commit. A separate pull request is recommended for the HTML rebuild.

### DIFF
--- a/01-backup.md
+++ b/01-backup.md
@@ -21,10 +21,9 @@ to keep track of what one person did and when.
 Even if you aren't collaborating with other people,
 version control is much better for this than this:
 
-<div>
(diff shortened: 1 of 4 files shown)


r-gaia-cs:
:+1:

abought:
Rebased this to address a merge conflict. Did the recent trainers workshop yield any estimates for the `minutes` field in place of the FIXME values?

iglpdc:
As it seems that we don't have good estimates for `minutes`, I'm merging this and adding a new issue with for the lesson duration.

Thanks @abought !

PR #43, opened 2015-02-12
remove color.ui settings

andreabedini:
As explained in the [git documentation](http://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#Colors-in-Git) the configuration variable `color.ui` is set to `auto` by default. There's no need to tell students to set this variable themselves.

### DIFF
--- a/01-backup.md
+++ b/01-backup.md
@@ -35,7 +35,6 @@ Here's how Dracula sets up his new laptop:
 ~~~ {.bash}
 $ git config --global user.name "Vlad Dracula"
 $ git config --global user.email "vlad@tran.sylvan.ia"
-$ git config --global color.ui "auto"
 $ git config --global core.editor "nano"
 ~~~
 


wking:
On Thu, Feb 12, 2015 at 02:42:32PM -0800, Andrea Bedini wrote:
> As explained in the [git
> documentation](http://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#Colors-in-Git)
> the configuration variable `color.ui` is set to `auto` by default.

That's just the Pro Git book (which is pretty good, but versioned
independently from the Git code and generally less detailed than [1]).
The Git docs say:

  $ git config --help
  ...
      color.ui
          ... Set it to always if you want all output not intended for
          machine consumption to use color, to true or auto (this is
          the default since Git 1.8.4) if you want such output to use
          color when written to the terminal.

I'm not sure how far back we want to support Git clients.  1.8.4 was
cut on 2013-08-23, but feel like we've had problems keeping Windows
and OS X users up to date with their clients.

[1]: http://git.kernel.org/cgit/git/git.git/tree/Documentation

andreabedini:
OSX now ships with git version 1.9.3, I guess Windows users can install the latest version of git (since they have to install git anyway). Anyway, I reckon missing on ui colors is acceptable compromise for who is stuck with an old version of git. In the end, the less things we configure the better.

wking:
On Thu, Feb 12, 2015 at 03:21:18PM -0800, Andrea Bedini wrote:
> OSX now ships with git version 1.9.3, ...

For folks who are running the current OS X ;).  Just because new
releases of OSes and software are available, doesn't mean that our
students show up with them installed.

> Anyway, I reckon missing on ui colors is acceptable compromise for
> who is stuck with an old version of git. In the end, the less things
> we configure the better.

I agree with the idea here, I'm just not sure how long our support
window should be, or how stale our students' systems are.  Our current
official minimum Git is 1.7.0 [1].  I'm fine with bumping that version
and merging this PR if we're comfortable cutting off folks with older
Gits (who as you point out, should be able to upgrade).  I'm also
comfortable leaving this as it stands for another year or two, to give
folks on old systems more time, since an extra line in the config
isn't too big a burden on everyone else.

[1]: https://github.com/wking/swc-setup-installation-test/blob/d13b26c9fb89beebe9ff9e9dfc86ba8ddb8d01cd/swc-installation-test-2.py#L704

jiffyclub:
I'd vote for leaving this as is for now. I can tell you from the trenches that we get plenty of people with ancient operating systems in our workshops.

Random thought: It could be an interesting extra challenge in `discussion.md` to have people experiment with turning `ui.color` on and off as practice using the config system.

andreabedini:
@jiffyclub there's little to none content about using the git configuration system in the lesson, I don't think we need to include practice on that. I'd like to reiterate the main point: the additional obscure configuration option is not worth the gain (the colour ui). People using old system can live without.

jiffyclub:
I disagree with your premise, it is absolutely worth turning on the color UI.

iglpdc:
I'd also vote for leaving it for now. I think it's not that much intrusive, allows us to support outdated OSes, and gives the instructor the chance to introduce, even briefly, the git config system. 

I'm closing it without merging for now. Thanks in any case, @andreabedini.

PR #44, opened 2015-02-12
Bring over Key Points from version 5.2

arohl:
Having the key points i.e. summary before the challenges is helpful to the learners I believe. I have made a small change to the key points by replacing the word "snapshot" which is not defined anywhere else with "revisions", which is defined and reused a few times. If want to introduce to snapshots, perhaps better placed in the collaboration section. Finally some formatting of my markdown needs to occur as am not sure what to put in {.xxxxxx} for key points

### DIFF
--- a/01-backup.md
+++ b/01-backup.md
@@ -804,6 +804,21 @@ $ git status --ignored
 nothing to commit, working directory clean
 ~~~
 
+> ## Key Points
+> *   Use `git config` to configure a user name, email address, editor, and other preferences once per machine.
+> *   `git init` initializes a repository.
+> *   `git status` shows the status of a repository.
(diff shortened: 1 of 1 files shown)


arohl:
Thanks @iglpdc - as you may have guessed I am new here ;-) Makes total sense, although looking at   `reference.md` I still think the use of the word *snapshot* is unnecessary - its not defined in the **glossary** but rather inline in the **A better kind of backup** section and I can't see how its definition is different to the definition of *revision* provided. So how do I close off this change request as I agree the change I suggested is inappropriate? I will then make a new one regarding 'reference.md`

iglpdc:
Yes, you're right. I'll close this, you send a new PR for the changes in the other file.

Thanks!

PR #36, opened 2015-02-06
A few extra notes for the introductory sections.

cryptarch:
### `index.md`
Propose:
* mention branching and merging.
* add a warning about non-text files (since questions about these come up a lot).
* add theses and blogs to list of things git is useful for.

### `01-backup.md`
Propose:
* mention `.gitconfig`, and that there are lots more options that can go in this file if people want to look them up.
* correct an instruction to highlight a line of plain text as `{.bash}`.
* since `git checkout` is used for switching to branches, previous description for this command is a little misleading.

PR #45, opened 2015-02-12
Point out the importance of tracking text files (part of ResBaz train-the-trainer SWC training)

mscook:
Also, while I do believe the PhD comic does get the point across to novices (and something they can all relate to), I do **not** believe that planting the seed that you could track changes to Word documents is an entirely good idea.

### DIFF
--- a/01-backup.md
+++ b/01-backup.md
@@ -804,7 +804,25 @@ $ git status --ignored
 nothing to commit, working directory clean
 ~~~
 
+
 ## Types of files that should be stored in Git
+
+Git works best tracking the changes to text files. Plain text files are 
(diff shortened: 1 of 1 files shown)


jiffyclub:
This has recently been covered via #36.

andreabedini:
to be more specific (and provided I understand the discussion in #36), the text @mscook is proposing to include is already present in [discussion.md](https://github.com/swcarpentry/git-novice/blob/gh-pages/discussion.md#non-text-files)

jiffyclub:
Exactly, thanks for clarifying.

iglpdc:
@mscook I'm closing this without merging, as most of it is included in #36.

PR #48, opened 2015-02-12
Update motivation.md

tremenyi:
As part of the train the trainer workshop, I am providing some input to this lesson.  
I have suggested some changes to the Why Git section.  
A correction of a typo and some additional reasons/motivations to use Git

### DIFF
--- a/motivation.md
+++ b/motivation.md
@@ -12,7 +12,11 @@ project.
 <section class="slide">
 ## Why Git?
 
-Because many people are using GitHub to collaborating.
+Because many people are using GitHub to collaborate.
+Git provides a platform to objectively share the problem solving process.  
+It facilitates reproducible research, providing a way to manage and record your workflow.  
+This process can enhance collaborators understanding of a probelm.  
+
 </section>
 <section class="slide">
 ## Next Steps
--- a/motivation.md
+++ b/motivation.md
@@ -12,10 +12,10 @@ project.
 <section class="slide">
 ## Why Git?
 
-Because many people are using GitHub to collaborate.
-Git provides a platform to objectively share the problem solving process.  
+GitHub provides a platform to objectively share the problem solving process.  
 It facilitates reproducible research, providing a way to manage and record your workflow.  
-This process can enhance collaborators understanding of a probelm.  
+This process can enhance collaborators' understanding of a probelm.  
+GitHub has a large, active and supportive community.  
 
 </section>
 <section class="slide">


wking:
On Thu, Feb 12, 2015 at 03:34:05PM -0800, Tomas Remenyi wrote:
> A correction of a typo…

Good catch :).

> … some additional reasons/motivations to use Git

This looks reasonable to me, but maybe we want more wording about
iterating on content?  It's currently a reading a bit to abstractly
with “the problem solving process” and “your workflow” where I'd
expect something more sepecific to version control.

Also “collaborators understanding” → “collaborators' understanding”.

motivation.md
```diff
@@ -12,7 +12,11 @@ project.
 <section class="slide">
 ## Why Git?
 
-Because many people are using GitHub to collaborating.
+Because many people are using GitHub to collaborate.
+Git provides a platform to objectively share the problem solving process.
```
jiffyclub: This line sounds kind of meaningless, I think the text would be fine without it.

jiffyclub:
Something to note is that `motivation.md` is meant to be converted to slides, so we want to avoid having whole paragraphs of text in them. They are meant to be an accompaniment to speaking, not something people will read on their own. (I think.)

wking:
On Thu, Feb 12, 2015 at 04:06:42PM -0800, Matt Davis wrote:
> Something to note is that `motivation.md` is meant to be converted
> to slides, so we want to avoid having whole paragraphs of text in
> them. They are meant to be an accompaniment to speaking, not
> something people will read on their own. (I think.)

Ah, it looks like they're rendered automatically as slides?  [1] has:

  motivation.md: slides for a short introductory presentation (three
    minutes or less) explaining what the lesson is about and why
    people would want to learn it. See "Introductory Slides" below.

and the 'slides' layout in [2].  Short and sweet is the way to go
then, although we might want some longer notes for instructor patter?

[1]: https://github.com/swcarpentry/lesson-template/blob/gh-pages/LAYOUT.md#pages
[2]: https://github.com/swcarpentry/lesson-template/blob/gh-pages/motivation.md

wking:
Here's the renderd slides for this lesson, to give folks a feel for
how much space is available [1].  I think what's there now is too
short, and the original version of this PR is too long.  What would be
just right?

[1]: http://swcarpentry.github.io/git-novice/motivation.html

jiffyclub:
See also the description at https://github.com/swcarpentry/lesson-template/blob/gh-pages/LAYOUT.md#motivational-slides. deck.js might have a way to embed speaker notes in the slides.

tremenyi:
The content I have is trying to be very conceptual, extremely entry level and answering the issues I have had learning what git is and why I should use it.  All up its about 5-10min of content.  
It doesnt really fit into each of the current pages, as its too simple compared to whats there already. 
I also have a concept map (draw on paper, photoed and cropped png) we developed during the workshop that I think helps too.

iglpdc:
I agree that the slides need some work. I don't think that what we have now describes well why we teach version control. I also don't like much slides cluttered with text, so I'd be awesome if we could add some presenter notes.

gvwilson:
Please have a look at 
http://swcarpentry.github.io/slideshows/introducing-software-carpentry/index.html 
and http://swcarpentry.github.io/slideshows/hosting-workshop/index.html 
for examples.
Thanks,
Greg

iglpdc:
I see. The slideshow that @gvwilson mentions has embedded the presenter notes. See:

https://github.com/swcarpentry/slideshows/blob/gh-pages/introducing-software-carpentry/index.html

Pretty cool, I have to say...

tremenyi:
Hi All, sorry I am super new to this, finding it hard to link all the ideas to which content needs what.
I'll revisit everything and figure out where best to put the content I have.  
One of mate calls 'commit' 'inflict', I;m sorry I might have done that... :-\

gvwilson:
On 2015-02-14 5:01 PM, Tomas Remenyi wrote:
>
> Hi All, sorry I am super new to this, finding it hard to link all the 
> ideas to which content needs what.
>
Hi Tomas,

Please don't apologize - this stuff is genuinely hard, and modifying 
someone else's lesson is at least as hard as modifying their code.

> I'll revisit everything and figure out where best to put the content I 
> have.
>
> One of mate calls 'commit' 'inflict', I;m sorry I might have done 
> that... :-\
>
Heh - I'm going to tweet that.

Thanks,
Greg

-- 
Dr. Greg Wilson    | gvwilson@software-carpentry.org
Software Carpentry | http://software-carpentry.org

PR #49, opened 2015-02-13
Rewrite and considerable fleshing out of motivation

hlapp:
Also moved the version clutter picture into the motivation. As a result, the `motivation.md` is no longer slides, but there wasn't much there anyway, and I'd argue these better not be slides.

### DIFF
--- a/01-backup.md
+++ b/01-backup.md
@@ -17,14 +17,8 @@ minutes: 60
 >     and explain why it is sometimes useful to do so.
 
 We'll start by exploring how version control can be used
-to keep track of what one person did and when.
-Even if you aren't collaborating with other people,
-version control is much better for this than this:
-
-<div>
-  <a href="http://www.phdcomics.com"><img src="fig/phd101212s.gif" alt="Piled Higher and Deeper by Jorge Cham, http://www.phdcomics.com" /></a>
-  <p>"Piled Higher and Deeper" by Jorge Cham, http://www.phdcomics.com</p>
-</div>
+to keep track of what one person did, when, and why, and why that is
+useful even if you aren't collaborating with other people.
 
 ## Setting Up
 
--- a/motivation.md
+++ b/motivation.md
@@ -21,6 +21,11 @@ time, keeps earlier versions around for posterity without burdening
 you with them, and represents a complete documented history of what
 you did at what time and why.**
 
+<div>
+  <a href="http://www.phdcomics.com"><img src="fig/phd101212s.gif" alt="Piled Higher and Deeper by Jorge Cham, http://www.phdcomics.com" /></a>
+  <p>How to avoid this for your code. "Piled Higher and Deeper" by Jorge Cham, http://www.phdcomics.com</p>
+</div>
+
 As a scientist, you always collaborate. Even if you carry out your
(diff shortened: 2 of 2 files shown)


iglpdc:
Thanks for your PR, @hlapp. 

All lessons should follow our lesson template, described [here](https://github.com/swcarpentry/lesson-template/blob/gh-pages/LAYOUT.md). This allows to automate much of the process and also reduces enormously the barrier for instructors to teach new materials. 

According to the template: 

> motivation.md: slides for a short introductory presentation (three minutes or less) explaining what the
> lesson is about and why people would want to learn it. See "Introductory Slides" below.

So, your changes cannot be merged here. I agree that the name motivation is a bit confusing, though. You could try to reuse part of your text to improve the topic lessons, or improve `motivation.md` sticking to the fact that they are meant to be slides.

hlapp:
That's fine if you decide not to use any of this. Just to give some background, my 3 PRs are from a [git lesson I taught based on the SWC material](https://github.com/Duke-GCB/GCB-Academy-2015-02-05/tree/master/git) as part of [a SWC-inspired workshop](http://duke-gcb.github.io/SciComp-Feb-2015/). The lesson received _very_ positive feedback. The PRs #49, #50, and #51 are my attempt at returning to you the changes that I felt were important for me to make to the material to make it more teachable.

I do think that in particular the template rule about a 3 minute slideshow for the motivation is misguided for this lesson. First, it shouldn't be a slideshow, but it needs to be narrated. The text I submitted here is just the basis for narration. Second, the Git section needs to be motivated well, and doing this well will need a little more than 3 minutes. A 3 minute slideshow as motivation for this lesson seems pointless to me. As I taught it, I erred on the other side, but I still think a thoroughly motivated argument will take substantially longer than 3 minutes. I also think that if the rest of the lesson is hands-on, if you start with slides you are subliminally messaging to the students that they don't need to be pay attention yet.

That said, these are just my opinions. All I can say is that these opinions come from trying to and actually teaching the material to a classroom of students none of whom was doing version control, and the vast majority of whom had never heard of the concept.

gvwilson:
The short motivational slideshows are meant to be ads for the lessons 
that learners (or people thinking of hosting a workshop) can watch on 
their own to get an idea of what we're going to be teaching - I wouldn't 
use them in the lesson itself.

hlapp:
So @gvwilson, what would then be the place for motivation material that you want to use as part of the lesson itself?

iglpdc:
I would love to have slides accompanied by some (hidden) presenter notes. In this case, in the event we need to use them in a live presentation, we have the corresponding text.

There's a related discussion in #48.

ethanwhite:
I'm late to the party on this, but I think that something along these lines would be beneficial at the beginning on the git lesson. As it is right now the lesson jumps almost immediately into the mechanics of git, and one of the big challenges with version control is explaining to students why they should do it and what the benefits are. 

@daisieh added material at her recent workshop [1] to the beginning of the first chunk of material in an attempt to accomplish this and I think it would be great to see some of what @hlapp and @daisieh have started developing built into the opening lesson in `01-backup.md`

[1] https://github.com/swcarpentry/git-novice/blob/2015-03-05-ubc-wis/01-backup.md

PR #50, opened 2015-02-13
Revamp of the open science part.

hlapp:
Some comments that are in part behind these changes:

1. I think it's a bad idea for a SWC lesson to list repository hosting providers. The list is bound to be incomplete, and thus involuntarily favoring some providers over others by listing some but not others, and that Github is a choice is self-evident from where the material is hosted. If anything, talk up the social coding features of Github and mention that it's not the only provider. I think making this a full-out discussion is beyond the scope of the material.
2. I don't see any merits from repeating descriptions and enumerations of licenses that are already, and more accurately made on the Creative Commons website. In particular, I don't see a reason to give the non-open CC licenses any room here (or, for that matter, airtime in the classroom). Those who don't understand the distinctions are, I would argue, best served with explaining what are appropriate choices and why. Contrasting with choices that aren't good ones for all kinds of reasons just clouds the message.
3. Singling out the GPL as a license that is different from all others is not a helpful Learning Objective. Rather, explaining how the infectiousness is a property that only the GPL has is a useful part of a Learning Objective on what license choices there are, and how one might choose between them.
4. That CC0 makes it unnecessary to cite the reused work is both false and a bad message to send. CC0 doesn't have a legal requirement for attribution, but the norms of citation are social norms, and they (fortunately!) don't go away just if one can't enforce it in a court of law anymore.

### DIFF
--- a/04-open.md
+++ b/04-open.md
@@ -5,10 +5,10 @@ subtitle: Open Science
 ---
 > ## Learning Objectives {.objectives}
 >
-> *   Explain how the GNU General Public License (GPL) differs from most other open licenses.
-> *   Explain the four kinds of restrictions that can be combined in a Creative Commons license.
-> *   Correctly add licensing and citation information to a project repository.
-> *   Outline options for hosting code and data and the pros and cons of each.
+> *   Explain how version control can be a leveraged as an electronic lab notebook for computational work.
+> *   Explain why adding licensing and citation information to a project repository is important.
+> *   Explain what license choices are available and how to choose one.
+> *   Explain how licensing and social expectations differ.
 
 > The opposite of "open" isn't "closed".
 > The opposite of "open" is "broken".
@@ -45,9 +45,10 @@ though,
 the process looks like this:
 
 *   The data that the scientist collects is stored in an open access repository
-    like [figshare](http://figshare.com/) or [Dryad](http://datadryad.org/)
-    as soon as it's collected,
-    and given its own DOI.
+    like [figshare](http://figshare.com/) or
+    [Zenodo](http://zenodo.org), possibly as soon as it's collected,
+    and given its own DOI. Or the data was already published and is stored in
+    [Dryad](http://datadryad.org/) 
 *   The scientist creates a new repository on GitHub to hold her work.
 *   As she does her analysis,
     she pushes changes to her scripts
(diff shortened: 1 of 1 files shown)


wking:
This makes sense to me, and I think the rewritten Licensing section is
moving in a good direction.  It needs a lot of copy-editing though,
and I'll hold of on that until we have a consensus on the general
thrust.

I'm more skeptical about removing the hosting section.  I think it's
useful to list major hosting providers (if only to help draw a
distinction between Git and GitHub).  It doesn't have to be complete,
but the current list hits the hosting sites and self-hosting options
I'd recommend to a newcomer.  It also touches on the restrictions,
costs, and benefits of the various approaches, which I think is
important.

Some minor issues and suggestions on the earlier changes:

* “can be a leveraged” → “can be leveraged”.
* You add trailing whitespace after the Dryad link (and maybe should
  end that line with a period, as the surrounding bullet-points do).
* “The benefits of version control…” is a fairly large sentence.  I'd
  prefer breaking it here: “… for your computational work.  In
  particular with …”.
* “Your stages of conception of your work…” → “The conceptual stages
  of your work…”.
* “These are stamped in a way that it is nearly impossible to forge.”
  is only true if you're both signing tags and versioning non-binary
  files.  And then it *still* only means that it's hard for *others*
  to edit your published history.  You can still massage your history,
  fake the timestamp, and re-sign the altered history, unless you have
  an independent third party *also* sign your published work.  Anyhow,
  I'd just drop this particular claim.
* “The version control repository is well amenable to archiving for
  perpetuity, and contains the entire history” → “The version control
  repository is easy to archive and contains the entire history”.

hlapp:
> I'm more skeptical about removing the hosting section.  I think it's useful to list major hosting providers (if only to help draw a distinction between Git and GitHub).

I don't disagree that it's useful. But having taught the lesson, I'm not sure how I would ever get there within 3-4 hours, even with non-novice students. To me, there'd always be things I would find more important to teach about Git and Github if I had time left, and so I'm not convinced about investing time into maintaining something that I'll never be able to get to in teaching. And this stuff about hosting is just bound to go out of date. Anywho, just my personal opinion.

wking:
On Fri, Feb 13, 2015 at 03:40:11PM -0800, Hilmar Lapp wrote:
> > I'm more skeptical about removing the hosting section.  I think
> > it's useful to list major hosting providers (if only to help draw
> > a distinction between Git and GitHub).
> 
> I don't disagree that it's useful. But having taught the lesson, I'm
> not sure how I would ever get there within 3-4 hours, even with
> non-novice students.

Good point.  I guess I'd rather move it to discussion.md, but you're
right that it doesn't belong in 04-open.md.

iglpdc:
I think the discussion about providers has to happen somewhere. We use GitHub during the whole lesson, but I think mentioning that you can have free private repos in other providers is worthy and help people that cannot or don't want to make their work public at this point. Maybe it's worth to considering to move it out of `04-open.md`.

04-open.md
```diff
@@ -69,74 +70,132 @@ the process looks like this:
 
 This open model accelerates discovery:
 the more open work is,
-[the more widely it is cited and re-used](http://www.plosone.org/article/info%3Adoi%2F10.1371%2Fjournal.pone.0000308).
+[the more widely it is cited and re-used](http://dx.doi.org/10.1371/journal.pone.0000308).
 However,
 people who want to work this way need to make some decisions
 about what exactly "open" means in practice.
 
+## Version control as electronic notebook
```
iglpdc: Maybe add "reproducible": "reproducible electronic notebook"?

04-open.md
```diff
@@ -69,74 +70,132 @@ the process looks like this:
...
+
+The benefits of version control in essence mean that, when used
+diligently, you can use version control as a form of electronic
+notebook for your computational work.
+
+* The conceptual stages of your work are documented, including who did
+  what and when. Every incremental step is stamped with an identifier
```
iglpdc: I'd simplify this sentence to:

"Every step is stamped with an identifier, the commit-id, which is essentially unique for most purposes."

04-open.md
```diff
@@ -69,74 +70,132 @@ the process looks like this:
...
-and include links to their full legal formulations.
-
-There is one other important license that doesn't fit into this categorization.
-Scientists (and other people) can choose to put material in the public domain,
-which is often abbreviated "PD".
-In this case,
-anyone can do anything they want with it,
-without needing to cite the original
-or restrict further re-use.
+At the latest when a repository with source code (and/or manuscript
```
iglpdc: I'd simplify to:

"... with source code, a manuscript, or other creative text..."

gvwilson:
We talk about hosting in 04-open.md

04-open.md
```diff
@@ -69,74 +70,132 @@ the process looks like this:
...
+and other creative writing text) becomes public, it should include a
+license, and a license disclaimer in prominent locations (such as the
```
iglpdc: This is confusing and I think is too detailed for the level we are aiming for. I'd like better something similar to the original formulation:

"every project have a file in its home directory redistributed as well called something like `LICENSE` or `LICENSE.txt`	that clearly states what the license is"

04-open.md
```diff
@@ -69,74 +70,132 @@ the process looks like this:
...
+top-level README, and at the top of source code files). This is because
+as creative works, source code is automatically eligible for
+intellectual property (and thus copyright) protection. Code that appears to be, or is expressly
+advertised as freely available has *not* waived or relinquished such
```
iglpdc: I'd remove "relinquished", "not waived" is clear enough for everyone.

04-open.md
```diff
@@ -69,74 +70,132 @@ the process looks like this:
...
+protection. Consequently, others who (re)use code that lacks a license
+statement do so enitrely on their own peril, because the author(s)
+of the software code can always unilaterally make such reuse illegal.
+
+A license solves this problem by granting rights to others (the
+licensees) that they would otherwise not have. What rights are being
+granted under which conditions differs, often only slightly, from one
+license to another. In contrast to proprietary licenses, the
+[licences](http://opensource.org/licenses/alphabetical) certified by
```
iglpdc: I'd change here "[licenses](...)" to "[open licenses](...)", even if redundant, to stress the contrast between open and proprietary.

04-open.md
```diff
@@ -69,74 +70,132 @@ the process looks like this:
...
-without needing to cite the original
-or restrict further re-use.
+At the latest when a repository with source code (and/or manuscript
+and other creative writing text) becomes public, it should include a
+license, and a license disclaimer in prominent locations (such as the
+top-level README, and at the top of source code files). This is because
+as creative works, source code is automatically eligible for
+intellectual property (and thus copyright) protection. Code that appears to be, or is expressly
+advertised as freely available has *not* waived or relinquished such
+protection. Consequently, others who (re)use code that lacks a license
```
iglpdc: I'd remove the sentence starting by "Consequently, ..." and modify merge it with the next one.

"A license allows the (re)use of code by others legally, by granting them rights that otherwise won't have"

04-open.md
hlapp: I meant to say _lab_ notebook. (It's not the notebook that's reproducible.)

04-open.md
hlapp: I actually think this sentence is important. Most people will not understand why using "free beer" code is dangerous if this isn't explained to them, and they will then also not understand why not stating a license for their public code is arguably worse than not publishing the code.

wking:
On Fri, Feb 13, 2015 at 05:05:46PM -0800, Greg Wilson wrote:
> We talk about hosting in 04-open.md

That's where it is now, but we @hlapp was making an argument for
removing it on time constraints [1], and putting it in discussion.md
sounds good to me [2].

[1]: https://github.com/swcarpentry/git-novice/pull/50#issuecomment-74346764
[2]: https://github.com/swcarpentry/git-novice/pull/50#issuecomment-74349003

gvwilson:
Thanks to @hlapp for the rewrite - thoughts below.

1.  I'd like to leave in explicit mention of Bitbucket and SourceForge to emphasize that Git isn't tied to GitHub (many or most learners think it is).
2.  And if we're taking out mentions of companies, we should take out Figshare, Zenodo, etc.  On balance, I think we're better with all of them.
3.  Everything we say is said elsewhere, but most of our learners won't ever go and read it.  I believe the overview of licensing (both CC and GPL/others) is therefore useful and should stay.
4.  Point #4 (citing CC0 work as a social norm) is a good one - thanks for the fix.

Thanks again,
Greg

daisieh:
I think I'm going to merge this one in, then make a new PR to re-add the hosting bit back. If you guys don't want that part added back, we can discuss it there.

hlapp:
Thanks @daisieh for stepping in here :smiley:

daisieh:
It's now part of my job, @hlapp :)

Issue #33, opened 2015-01-25
Challenge at end of 01-backup.md - git repo inside a git repo


willingc:
I'm wondering if there has been feedback from workshops on the last challenge of a git repo inside a git repo. While I like the challenge question of why it is a bad idea, I think that the steps are confusing for someone learning git. New learners of git (especially if they are learning bash too) are easily confused about where they are in the filesystem tree.

Recommendation: Remove the steps in this challenge and keep the question for students to ponder. Add a challenge that has the student moving through the history of the log using HEAD~1 would be a better exercise.

jhamrick:
I agree, I think the steps in this challenge will be confusing to some students, especially if they are new to the command line.

> Add a challenge that has the student moving through the history of the log using HEAD~1 would be a better exercise.

I'm not entirely sure how this is related to a git repo inside a git repo -- could you clarify? Or are you just suggesting adding a different unrelated challenge?

willingc:
@jhamrick different unrelated challenge. Merely an exercise to get users to understand 'HEAD' and how ~1 is syntactic sugar for moving back 1 from HEAD. For example, git show HEAD~2 displays the commit that is two before the HEAD.

iglpdc:
I think that the challenge is fine, although it's true that the wording may be confusing. It happens quite often that learners are overwhelmed at the beginning of the lesson (they have to deal not only with `git`, but also with the command line) and some `git init` in the wrong place, either a repo inside a repo, or their entire home folder, or something similar.

Despite the fix is pretty easy, they aren't able to do it yet. I like the challenge because it's a very good moment to reinforce the idea that Git is folder-based.

I'd vote to keep the challenge, but rewording it to make it closer to the reality of the lesson. For example, `alpha` can be `planets` and `beta` could be `paper`, an independent repo for the paper associated to the planet research, or `moons`, a new repo on moon research. I'd also add a question on how to fix it, and stress that, in general, that's the way of fixing wrong `git init`'s. 

Maybe we should add also a corresponding note in the instructor guide.

willingc:
@iglpdc I like your comments. If the challenge can be reworded to be more beginner friendly as you suggest, I agree that it would be helpful to reinforce the idea that Git is folder based. Do you wish to try to reword the challenge as you suggested? :smiley_cat:

iglpdc:
Ok, @willingc, I'll send a PR tonight rephrasing the challenge.

iglpdc:
Closing by #66.

[...]

PR #54, opened 2015-02-18
Rephrase the last challenge. Add question on how to undo a git-init.

iglpdc:
Addresses the enhancement proposed in #33.

It makes the last challenge (why is bad to create a repo inside a repo) to follow a bit more the lesson theme. More importantly, it adds a question on what to do when you create a repo in the wrong place. This is something that happens quite commonly during this workshops, as learners, who are not yet proficient with the command line, put under version control the wrong folder, e.g. their whole home directory.


gvwilson:
How *would* you undo the second `git init`?  `rm -rf .git`?

iglpdc:
Yes, `rm -rf .git` is what I was thinking. 

I find often learners that had put all their Desktop/home dir under version control and they only realize when they do a `git status` further in the lesson. As `rm -rf` is risky, specially if you do it around one of these places, I usually go to their laptop to double check they aren't removing something they don't want. Doing this at the beginning of the class, in a more controlled way could save some time later in the lesson and also reinforce the idea of what `git init` actually doing (just creating `.git`, from a simplistic point of view) and that all Git magic is self-contained in this directory.

gvwilson:
How about `mv .git /tmp/this-was-the-git-folder` and then `cd /tmp; rm 
-rf this-was-the-git-folder` with an explanation of why?

iglpdc:
Ok, I've added a short paragraph about this to `instructors.md`

iglpdc:
Could someone else take a look at this?

gvwilson:
@daisieh can you please look at this one?

daisieh:
Why not have the challenge itself be "accidentally create a git repo out of your home directory?" Then the warning you're making becomes more explicit.

I'd also recommend, especially for newbies to the command line, that the way to recover is what @gvwilson said: `mv .git tempgit` (it doesn't even have to be out of the directory, just changed to something visible) and then they can look at what the git directory looks like a little before deleting it, just like any other directory.

iglpdc:
Oh, sorry, guys. I did the changes @gvwilson mentioned, but committed them to another branch, that's why I was asking you to review it again... I'll fix it and then you can take a look.

iglpdc:
@daisieh I don't like the idea of making people put their home directory under git and then asking them to use `rm -rf` there: eventually, someone is going to make a mistake and accidentally delete their home dir... 

Also we want to say that making a repo out of your home or your Desktop is not wrong (probably not very useful either), but making repos inside repos is.

daisieh:
Well, not that you should do it, but it could be a thought exercise: like, what would happen if you did it?

instructors.md
```diff
@@ -3,28 +3,171 @@ layout: page
...
+    arise naturally once we start talking about
+    using public services like GitHub to store files;
+    and
+2.  the discussion gives learners a chance to catch their breath
+    after what is often a frustrating couple of hours.
 
-## [A Better Kind of Backup](01-backup.html)
+## Teaching Notes
+
+<<<<<<< HEAD
```
daisieh: a merge conflict note in the PR??

daisieh:
@iglpdc, are there supposed to be this many files committed in this PR?

iglpdc:
This is messed up. I'm closing this without merging and adding the changes again.

PR #55, opened 2015-02-21
Remove term *snapshot* from `reference.md`

arohl:
the word *snapshot* is not defined anywhere in the lesson but does get used in `reference.md`. Rather than define *snapshot*, I have changed the text to used the defined term *revision*.  Another approach could be to use the term *change set* which is defined in the Glossary but doesn't appear anywhere else in the lesson.

### DIFF
--- a/reference.md
+++ b/reference.md
@@ -11,9 +11,9 @@ subtitle: Reference
 *   `git status` shows the status of a repository.
 *   Files can be stored in a project's working directory (which users see),
     the staging area (where the next commit is being built up)
-    and the local repository (where snapshots are permanently recorded).
+    and the local repository (where revisions are permanently recorded).
 *   `git add` puts files in the staging area.
-*   `git commit` creates a snapshot of the staging area in the local repository.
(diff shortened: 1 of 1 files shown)


reference.md
```diff
@@ -11,9 +11,9 @@ subtitle: Reference
 *   `git status` shows the status of a repository.
 *   Files can be stored in a project's working directory (which users see),
     the staging area (where the next commit is being built up)
-    and the local repository (where snapshots are permanently recorded).
+    and the local repository (where revisions are permanently recorded).
 *   `git add` puts files in the staging area.
-*   `git commit` creates a snapshot of the staging area in the local repository.
+*   `git commit` save the revisions in the staging area to the local repository.
```
r-gaia-cs: " `git commit` save" -> "`git commit` saves"

r-gaia-cs:
:+1: to merge after address my comment.

arohl:
Agree with @r-gaia-cs fix

iglpdc:
Yes, fix this word and we can merge.

Issue #39, opened 2015-02-12
Add instructions for configuring commonly-used editors to work with Git


gvwilson:
Originally https://github.com/swcarpentry/bc/issues/600: we should have instructions somewhere explaining how to configure commonly-used editors to work with Git on all three platforms.

iglpdc:
Closed by #56.

PR #56, opened 2015-02-22
Adds configuration for commonly used text editors

iglpdc:
Adds a table in `01-backup.md` with proper `git config` for the following editors: nano, Sublime Text 2 (Win & Mac), Kate, Text Wrangler, and Notepad++. Not the prettiest thing, but I don't find any other way to have them all.  Fixes issue #39.

(I found the list in some notes I snatched long ago from one @karthik 's repos :-) )

I don't have access to a Windows machine: can someone else check the two Windows beauties? Do they need to modify the Windows path to get it working?

### DIFF
--- a/01-backup.md
+++ b/01-backup.md
@@ -35,12 +35,20 @@ Here's how Dracula sets up his new laptop:
 $ git config --global user.name "Vlad Dracula"
 $ git config --global user.email "vlad@tran.sylvan.ia"
 $ git config --global color.ui "auto"
-$ git config --global core.editor "nano"
 ~~~
 
-(Please use your own name and email address instead of Dracula's,
-and please make sure you choose an editor that's actually on your system,
-such as `notepad` on Windows.)
+(Please use your own name and email address instead of Dracula's.)
+
+He also has to configure his favorite text editor, following this table:
+
+| Editor             | Configuration command                            |
+|:-------------------|:-------------------------------------------------|
+| nano               | `$ git config --global core.editor "nano"`       |
+| Sublime Text (Mac) | `$ git config --global core.editor "subl -n -w"` |
+| Text Wrangler      | `$ git config --global core.editor "edit -w"`    |
+| Sublime Text (Win) | `$ git config --global core.editor "'c:/program files/sublime text 2/sublime_text.exe' -w"` |
+| Notepad++          | `$ git config --global core.editor "'c:/program files (x86)/Notepad++/notepad++.exe' -multiInst -notabbar -nosession -noPlugin"`|
+| Kate (Linux)       | `$ git config --global core.editor "kate"`       |
 
 Git commands are written `git verb`,
 where `verb` is what we actually want it to do.
--- a/01-backup.md
+++ b/01-backup.md
@@ -47,8 +47,9 @@ He also has to set his favorite text editor, following this table:
 | Text Wrangler      | `$ git config --global core.editor "edit -w"`    |
 | Sublime Text (Mac) | `$ git config --global core.editor "subl -n -w"` |
 | Sublime Text (Win) | `$ git config --global core.editor "'c:/program files/sublime text 2/sublime_text.exe' -w"` |
-| Notepad++          | `$ git config --global core.editor "'c:/program files (x86)/Notepad++/notepad++.exe' -multiInst -notabbar -nosession -noPlugin"`|
-| Kate               | `$ git config --global core.editor "kate"`       |
+| Notepad++ (Win)    | `$ git config --global core.editor "'c:/program files (x86)/Notepad++/notepad++.exe' -multiInst -notabbar -nosession -noPlugin"`|
+| Kate (Linux)       | `$ git config --global core.editor "kate"`       |
+| Gedit              | `$ git config --global core.editor "gedit -s"`   |
 
 Git commands are written `git verb`,
 where `verb` is what we actually want it to do.
--- a/01-backup.md
+++ b/01-backup.md
@@ -39,7 +39,7 @@ $ git config --global color.ui "auto"
 
 (Please use your own name and email address instead of Dracula's.)
 
-He also has to configure his favorite text editor, following this table:
+He also has to set his favorite text editor, following this table:
 
 | Editor             | Configuration command                            |
 |:-------------------|:-------------------------------------------------|
@@ -62,6 +62,12 @@ we're telling Git:
 
 The four commands above only need to be run once:
 the flag `--global` tells Git to use the settings for every project on this machine.
+
+You can check your settings at any time:
+
+~~~ {.bash}
+$ git config --list
+~~~
 
 > ## Proxy {.callout}
 >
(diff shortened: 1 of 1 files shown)


01-backup.md
```diff
@@ -35,12 +35,20 @@ Here's how Dracula sets up his new laptop:
...
+(Please use your own name and email address instead of Dracula's.)
+
+He also has to set his favorite text editor, following this table:
+
+| Editor             | Configuration command                            |
+|:-------------------|:-------------------------------------------------|
+| nano               | `$ git config --global core.editor "nano"`       |
+| Sublime Text (Mac) | `$ git config --global core.editor "subl -n -w"` |
+| Text Wrangler      | `$ git config --global core.editor "edit -w"`    |
+| Sublime Text (Win) | `$ git config --global core.editor "'c:/program files/sublime text 2/sublime_text.exe' -w"` |
```
gvwilson: Why not put the two Sublime Text lines together?

01-backup.md
```diff
@@ -35,12 +35,20 @@ Here's how Dracula sets up his new laptop:
...
+| Notepad++          | `$ git config --global core.editor "'c:/program files (x86)/Notepad++/notepad++.exe' -multiInst -notabbar -nosession -noPlugin"`|
+| Kate (Linux)       | `$ git config --global core.editor "kate"`       |
```
gvwilson: I don't think this needs `(Linux)` - I don't know anywhere else Kate is ever used.

gvwilson:
A couple of minor suggestions, but this is a great addition.

01-backup.md
iglpdc: I thought that maybe some people will enter the wrong line for the OS or even both, but probably is OK to put them together.

iglpdc:
I'll merge as soon as someone with Windows confirms the line for Notepad++ is correct.

01-backup.md
wking: On Sun, Feb 22, 2015 at 07:26:29AM -0800, Greg Wilson wrote:
> > +| nano               | `$ git config --global core.editor "nano"`       |

`nano -w` is probably a better idea, especially when we're likely to
be handling a whitespace-sensitive languages like Python.

wking:
On Sun, Feb 22, 2015 at 08:08:07AM -0800, Ivan Gonzalez wrote:
> I'll merge as soon as someone with Windows confirms the line for
> Notepad++ is correct.

I don't have access to a Windows box either, but in general "program
files" vs. "program files (x86)" depends on the architecture that the
program was compiled for (x86 or amd64).  For some more details see
swcarpentry/windows-installer@23d9af6e (swc-windows-installer.py:
Search ProgramW6432 and ProgramFiles(x86), 2014-07-09).  Do we know
that there are only 64-bit builds of Sublime and only 32-bit builds of
Notepad++?

iglpdc:
Thanks, @wking. It seems that Sublime has both 64- and 32-bit [1], although in our current setup instructions we offer Sublime as an option only for Mac. (Maybe we should add the Windows version?)

Regarding notepad++, I have no idea: it's not clear to me from their website [2].

[1] http://www.sublimetext.com/2
[2] http://notepad-plus-plus.org/

01-backup.md
```diff
@@ -35,12 +35,20 @@ Here's how Dracula sets up his new laptop:
...
+
+He also has to set his favorite text editor, following this table:
+
+| Editor             | Configuration command                            |
+|:-------------------|:-------------------------------------------------|
+| nano               | `$ git config --global core.editor "nano -w"`    |
+| Text Wrangler      | `$ git config --global core.editor "edit -w"`    |
+| Sublime Text (Mac) | `$ git config --global core.editor "subl -n -w"` |
+| Sublime Text (Win) | `$ git config --global core.editor "'c:/program files/sublime text 2/sublime_text.exe' -w"` |
+| Notepad++          | `$ git config --global core.editor "'c:/program files (x86)/Notepad++/notepad++.exe' -multiInst -notabbar -nosession -noPlugin"`|
```
r-gaia-cs: I would like to add "(Win)".

01-backup.md
```diff
@@ -35,12 +35,20 @@ Here's how Dracula sets up his new laptop:
...
+| Kate               | `$ git config --global core.editor "kate"`       |
```
r-gaia-cs: I would like to add "(*nix)" since I believe that you can run Kate on Windows.

01-backup.md
```diff
@@ -35,12 +35,20 @@ Here's how Dracula sets up his new laptop:
...
 ~~~
 
-(Please use your own name and email address instead of Dracula's,
-and please make sure you choose an editor that's actually on your system,
-such as `notepad` on Windows.)
+(Please use your own name and email address instead of Dracula's.)
+
+He also has to set his favorite text editor, following this table:
+
+| Editor             | Configuration command                            |
```
r-gaia-cs: Add Gedit? `$ git config --global core.editor "gedit -s"`.

01-backup.md
gvwilson: > I would like to add "(*nix)" since I believe that you can run Kate on 
> Windows.
>
Let's just say "Unix" please - '*nix*' doesn't match Linux anyway.

iglpdc:
I've added again the OS for `kate` and `notepad++`. I think it's a bit redundant though, because it's very unlikely that someone uses them outside Linux and Win.

01-backup.md
```diff
@@ -35,12 +35,21 @@ Here's how Dracula sets up his new laptop:
...
+|:-------------------|:-------------------------------------------------|
+| nano               | `$ git config --global core.editor "nano -w"`    |
+| Text Wrangler      | `$ git config --global core.editor "edit -w"`    |
+| Sublime Text (Mac) | `$ git config --global core.editor "subl -n -w"` |
+| Sublime Text (Win) | `$ git config --global core.editor "'c:/program files/sublime text 2/sublime_text.exe' -w"` |
+| Notepad++ (Win)    | `$ git config --global core.editor "'c:/program files (x86)/Notepad++/notepad++.exe' -multiInst -notabbar -nosession -noPlugin"`|
+| Kate (Linux)       | `$ git config --global core.editor "kate"`       |
+| Gedit              | `$ git config --global core.editor "gedit -s"`   |
```
r-gaia-cs: Also add "(Linux)"?

01-backup.md
```diff
@@ -35,12 +35,21 @@ Here's how Dracula sets up his new laptop:
...
-(Please use your own name and email address instead of Dracula's,
-and please make sure you choose an editor that's actually on your system,
-such as `notepad` on Windows.)
+(Please use your own name and email address instead of Dracula's.)
+
+He also has to set his favorite text editor, following this table:
+
+| Editor             | Configuration command                            |
+|:-------------------|:-------------------------------------------------|
+| nano               | `$ git config --global core.editor "nano -w"`    |
```
r-gaia-cs: This works on Windows?

r-gaia-cs:
> I've added again the OS for kate and notepad++.

Thanks.

> I think it's a bit redundant though, because it's very unlikely that someone uses them outside Linux and Win.

In this case I prefer make it very clear to avoid problems.

01-backup.md
wking: On Mon, Feb 23, 2015 at 05:26:30PM -0800, Raniere Silva wrote:
> > +| nano               | `$ git config --global core.editor "nano -w"`    |
> 
> This works on Windows?

Yup, because our Windows installer installs nano [1] and puts it in
the PATH [2,3].  It looks like nano is installed on OS X too [4], and
it's installed by default on most Linux distributions I've seen.

[1]: https://github.com/swcarpentry/windows-installer/blob/v0.2/swc-windows-installer.py#L284
[2]: https://github.com/swcarpentry/windows-installer/blob/v0.2/swc-windows-installer.py#L287
[3]: https://github.com/swcarpentry/windows-installer/blob/v0.2/swc-windows-installer.py#L291
[4]: https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/nano.1.html

PR #61, opened 2015-03-03
Added teaching note for teaching collaboration and conflict lessons

ttimbers:


wking:
On Tue, Mar 03, 2015 at 02:11:01PM -0800, Tiffany Timbers wrote:
>   * Added teaching note for teaching collaboration and conflict lessons

I think the rest of the file is using spaces for indentation instead
of the tabs used here, but +1 on the content.  I'm smiling just
imaging dramatic conflict resolution… :).

gvwilson:
I'll fix the spaces-tabs thing after merging - thanks very much for the 
note.

PR #62, opened 2015-03-08
Lessons from WiSE workshop at UBC, March 5-6, 2015

daisieh:
NB: This pull request isn't supposed to be for direct merging into the master branch: this is just to leave a record of the lessons @ttimbers and I taught at this workshop, so we can share and discuss...

The main changes: I stripped out a lot of content from `01-backup` and added some introductory concept material as well. I added a short description of detached HEAD and master to `02-collab`. @ttimbers modified `02-collab` and `03-conflict` into an interactive lesson, `role-play`. We ended the second day with a lesson I added, `forking`, which introduces students to forking and pull requests.


role-play.md
```diff
@@ -0,0 +1,334 @@
+---
```
ahmadia: This is really well thought-through and presented.

daisieh:
PR #64 is the same request, only merged against a different branch in swcarpentry, not the main gh-pages branch.

daisieh:
Comments can still be made, but this PR doesn't need merging.

PR #66, opened 2015-03-12
Reword repo inside repo challenge

iglpdc:
Addresses the enhancement proposed in #33. See also the discussion on #54.

It makes the last challenge (why is bad to create a repo inside a repo) to follow a bit more the lesson theme. More importantly, it adds a question on what to do when you create a repo in the wrong place. This is something that happens quite commonly during this workshops, as learners, who are not yet proficient with the command line, put under version control the wrong folder, e.g. their whole home directory.

### DIFF
--- a/01-backup.md
+++ b/01-backup.md
@@ -856,16 +856,19 @@ nothing to commit, working directory clean
 
 > ## Places to Create Git Repositories {.challenge}
 >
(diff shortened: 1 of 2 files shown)


daisieh:
This looks good to me, merging.

PR #65, opened 2015-03-12
undo deletion of hosting section

daisieh:
Adding back the deleted hosting section from 04-open.md

(deleted in PR #50)

PR #67, opened 2015-03-12
Alternative history for restoring hosting

wking:
This is #65 but with a merge to preserve 'git blame ...' information.


daisieh:
Mergey merge!

PR #68, opened 2015-03-14
Update html to match markdown

ethanwhite:
The html files had gotten out of sync with the markdown. This updates them
by running 'make preview'.


iglpdc:
Thanks!

PR #58, opened 2015-02-22
Breaks up 01-backup.md in several topics

iglpdc:
Breaks the first topic of the lesson, `01-backup.md`, in 5 different topics, roughly corresponding to the sections in the previous version of `01-backup.md`. We need this to comply with the lesson template and to avoid having really long, unstructured topics. The new files are:

- `01-setup.md`: on how to setup your laptop for `git`,
- `02-create.md`: on how to create a repo with `git init`,
- `03-changes.md`: on the cycle of making changes, adding, and committing,
- `04-history.md`: on exploring history and restoring older versions,
- `05-ignore.md`: on how to use `.gitignore`.

Challenges have been moved to the end of the corresponding topic. The contents of each topic are not changed from the previous version. The rest of the topics (`collab`, `conflict`, and `open`) are renumbered appropriately. Some changes in other files are included to adapt to the new filenames. New time estimates are provided.

Fixes #29 .

PR #69, opened 2015-03-14
Minor language simplification

ethanwhite:
Language like "do so" and "for posterity" may be confusing for non-native
English speakers. This makes this language more straight forward.

This PR is currently missing updated html because until #68 goes in the diff shows far more change that has actually happened. I'll be happy to add the html changes after #68 is merged.

### DIFF
--- a/03-changes.md
+++ b/03-changes.md
@@ -60,7 +60,7 @@ nothing added to commit but untracked files present (use "git add" to track)
 
 The "untracked files" message means that there's a file in the directory
 that Git isn't keeping track of.
-We can tell Git that it should do so using `git add`:
+We can tell Git to track a file using `git add`:
 
 ~~~ {.bash}
 $ git add mars.txt
@@ -84,7 +84,7 @@ $ git status
 ~~~
 
 Git now knows that it's supposed to keep track of `mars.txt`,
-but it hasn't yet recorded any changes for posterity as a commit.
+but it hasn't recorded these changes as a commit yet.
 To get it to do that,
 we need to run one more command:
 


r-gaia-cs:
+1 to merge.

wking:
On Sat, Mar 14, 2015 at 12:59:33PM -0700, Ethan White wrote:
>   * Minor language simplification

Looks good to me.

wking:
On Sat, Mar 14, 2015 at 04:01:02PM -0700, W. Trevor King wrote:
> Sat, Mar 14, 2015 at 12:59:33PM -0700, Ethan White:
> >   * Minor language simplification
>
> Looks good to me.

Actually, we may want to keep the “as a commit” qualifier, since Git
*does* record the file at ‘git add …’ time:

  $ git init
  Initialized empty Git repository in /tmp/a/.git/
  $ echo stuff > mars.txt
  $ git add mars.txt
  $ rm mars.txt
  rm: remove regular file ‘mars.txt’? y
  removed ‘mars.txt’
  $ git --no-pager status
  On branch master

  Initial commit

  Changes to be committed:
    (use "git rm --cached <file>..." to unstage)

          new file:   mars.txt

  Changes not staged for commit:
    (use "git add/rm <file>..." to update what will be committed)
  	  (use "git checkout -- <file>..." to discard changes in working directory)

          deleted:    mars.txt

  $ git checkout -- mars.txt
  $ cat mars.txt
  stuff

It's just that without a commit, doing something like checking out
another branch will clobber the index (unless you stashed it).  After
the index is removed, the added mars.txt blob (and the tree containing
it) will be dangling, and will eventually be cleaned up by Git's
garbage collection.  We certainly don't want to get into all of that
with novices, but I don't think we want an unqualified “hasn't
recorded these changes yet” either.

So I'd suggest:

  … but it hasn't recorded these changes as a commit yet.

iglpdc:
Thanks @ethanwhite ! I agree with @wking that keeping "as commit" would be clearer: "... these changes as a commit yet".

daisieh:
@ethanwhite: I merged the big reorg PR #58, so please refactor the change against that new file.

ethanwhite:
I've refactored this against #58 and made the fix suggested by @wking.

PR #70, opened 2015-03-14
Improve commit messages in 01-backup

ethanwhite:
We probably don't want to get into the details of good commit message style
in this lesson, but we should try to model it as best as possible. This updates
the commit messages to make them a little better/more standard.

This PR is currently missing updated html because until #68 goes in the diff shows far more change that have actually happened. I'll be happy to add the html changes after #68 is merged.

### DIFF
--- a/03-changes.md
+++ b/03-changes.md
@@ -89,10 +89,10 @@ To get it to do that,
 we need to run one more command:
 
 ~~~ {.bash}
-$ git commit -m "Starting to think about Mars"
+$ git commit -m "Start notes on Mars as a base"
 ~~~
 ~~~ {.output}
-[master (root-commit) f22b25e] Starting to think about Mars
+[master (root-commit) f22b25e] Start notes on Mars as a base
  1 file changed, 1 insertion(+)
  create mode 100644 mars.txt
 ~~~
@@ -132,7 +132,7 @@ commit f22b25e3233b4645dabd0d81e651fe074bd8e73b
 Author: Vlad Dracula <vlad@tran.sylvan.ia>
 Date:   Thu Aug 22 09:51:46 2013 -0400
 
-    Starting to think about Mars
+    Start notes on Mars as a base
 ~~~
 
 `git log` lists all revisions  made to a repository in reverse chronological order.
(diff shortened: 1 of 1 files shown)


r-gaia-cs:
+1 to merge.

wking:
On Sat, Mar 14, 2015 at 01:06:49PM -0700, Ethan White wrote:
> We probably don't want to get into the details of good commit
> message style in this lesson, but we should try to model it as best
> as possible. This updates the commit messages to make them a little
> better/more standard.

This change:

  - The two moons may be a problem for Wolfman
  + Add concerns about effects of Mars' moons on Wolfman

is in the *content* of mars.txt, not the commit message.

You take “Thoughts about the climate” to “Add discussion of Mars'
climate”.  I like yours more than the original, but how about “Discuss
Mars' climate for Mummy”?  That will distinguish the commit from
subsequent commits which may also discuss the climate.  If it was my
moon-base project, I'd like “mars: Discuss climate for Mummy”, but I
think the non-prefixed sentence may be less confusing for novices.

Regardless of the chosen climate commit message, I think this change
is good and moves us in the right direction (although I don't think we
want the content change for mars.txt).

> This PR is currently missing updated html because until #68 goes in
> the diff shows far more change that have actually happened. I'll be
> happy to add the html changes after #68 is merged.

The maintainers are supposed to handle updating the HTML anyway, to
avoid conflicts in autogenerated-content from folks submitting HTML
updates in parallel PRs.  See swcarpentry/lesson-template#159, and
point 3 of [1] which came from swcarpentry/lesson-template#170.

[1]: https://github.com/swcarpentry/lesson-template/blob/af883468d2058f7af56938cafa4ae4b33ddb5fc4/CONTRIBUTING.md#getting-started

iglpdc:
Yes, we don't need to change the contents of the file. @ethanwhite, could you fix that and then I'll merge?

By the way, I see that we don't give any explanation on how to write a commit message. Without going into much details, I think one good explanatory sentence (together with an entry in the lesson glossary) would be nice. I'm opening a new issue with this.

daisieh:
@ethanwhite: I merged the big reorg PR #58, so please refactor the change against that new file.

ethanwhite:
Thanks for the great review @wking. I had accidentally find and replaced the wrong text for the 'moons' commit - thanks for catching that. I have now:

1. Re-rolled this against #58
2. Fixed the accidental replacement of the file content by changing the associated commit message instead
3. Improved the climate commit message building on @wking's suggestion.

wking:
f5009ba looks good to me :).

PR #73, opened 2015-03-17
remove old versions

daisieh:


daisieh:
Need to remove the pre-rebased versions of 01-backup.* and the old html files.

PR #81, opened 2015-03-22
Adapt instructors.md to the lesson template

### LINKED ISSUE #79: Move section on Gedit out of instructors' notes
The section on Gedit in the instructors' guide seems out of place - it should be moved into discussion of other editors.

iglpdc:
The use of "gedit -s" (equivalent to "--standalone") is already want we recommend in `01-setup.md` and the explanation of why is not needed here. Also, the previous structure didn't comply with our lesson template. This removes the explanation from this file (it's not added anywhere else either.)

This fixes #79.


iglpdc:
I've added the standard headers for "Legend" and "Overall", as well as one per section topic in the current lesson layout. I've also moved some of the notes under "Teaching notes" to the proper section, when the point was relevant to that particular part of the lesson.

daisieh:
Looks fine to me. Merging.

PR #91, opened 2015-04-07
code/backup.sh: Remove this unused file

wking:
It doesn't look like we've ever referenced this file (says
`git log -p -G backup.sh`).


iglpdc:
Good. Does anybody has some use for this file? I was thinking (now that I know it exists) that it may be useful for posting stuff in the Etherpad or something along these lines.

Otherwise, I'll merge this.

gvwilson:
I used to use this to check that all the steps I went through in the lesson would actually work.  The lesson has moved on, so +1 for deleting this file.

iglpdc:
Great! Let's delete it then.

Issue #82, opened 2015-03-23
Add end of line configuration in 01-setup.md


iglpdc:
In the teaching notes in `instructors.md`, we mention the problem of end of line characters for learners using Windows. I think this could be solved during the configuration in `01-setup.md` by setting:

    git config --global core.autocrlf input 

for Mac OS and Linux, and,

    git config --global core.autocrlf true

for Windows, [as recommended by Github](https://help.github.com/articles/dealing-with-line-endings/#platform-all).

Shouldn't that fix all end-of-line issues for everyone? It's just one more line and a little bit harder to explain than the rest, but may be worthy.

wking:
On Sun, Mar 22, 2015 at 09:01:51PM -0700, Ivan Gonzalez wrote:

> In the teaching notes in `instructors.md`, we mention the problem of
> end of line characters for learners using Windows. I think this
> could be solved during the configuration in `01-setup.md` by
> setting:
> …
>     git config --global core.autocrlf true
> 
> for Windows. 

You might also want:

  $ git config --global core.safecrlf warn

There's a more extensive discussion of this setting in git-config(1).

Personally, I prefer leaving these settings alone and using a
reasonable text editor.  Emacs and nano both definately handle
auto-detecting the end-of-line character(s), so users can edit the
files transparently regardless of the endlines it uses.  I expect
other serious editors do the same (I vaguely remember Notepad++ being
able to handle LF line endings, as well as Window's default CRLF).
Are we suggesting editors that *can't* handle that?  The only other
reason I can think of for using these settings is if we don't want to
pollute a repository with non-native line endings, and I don't think
we care that much about such minor polishing for the novice lesson.

daisieh:
I'd like to leave this in the instructor notes: weird config issues that only affect a subset of users and require unnecessary explanations should be saved to explain to just the few people who need it, I think.

daisieh:
My concern is that, as it is, the git config global things we already tell them at the very beginning set up this atmosphere of "magic incantations" about Git, and I'd like to minimize that until they feel a little more grounded in it. I think dealing with adding config statements as needed during training works better for that. But I do think we should have this page easily accessible by the instructors. I'll make a PR for that.

iglpdc:
I'm closing this by #92.

PR #92, opened 2015-04-17
Link to Github line ending page

daisieh:
Add a link to Github’s suggested config for dealing with line endings. Addresses Issue #82

### DIFF
--- a/instructors.md
+++ b/instructors.md
@@ -86,6 +86,7 @@ We close with material on licensing because:
     different editors may or may not add a newline to the last line of a file.)
     Take a moment to explain these issues,
     since learners will almost certainly trip over them again.
+    If learners are running into line ending problems, Github has a [page](https://help.github.com/articles/dealing-with-line-endings/#platform-all) that helps with troubleshooting.
 
 *   We don't use a Git GUI in these notes because
     we haven't found one that installs easily and runs reliably on the three major operating systems,


iglpdc:
Good. Are you adding the html for any particular reason? Otherwise, we should drop it for the commit and then update it later.

daisieh:
Just habit; I did `make preview` to verify the change and then committed the html as well as the md change. Is it hurting anything?

wking:
On Tue, Apr 21, 2015 at 12:14:47PM -0700, Daisie Huang wrote:
> Just habit; I did `make preview` to verify the change and then
> committed the html as well as the md change. Is it hurting anything?

Not if it doesn't cause merge conflicts.  But since you just have one
commit, it's reasonably straightforward to drop the HTML changes.
Something like:

  $ git checkout line-endings  # get on your feature branch
  $ git checkout HEAD^ -- instructors.html  # clobber with the parent's HTML
  $ git commit -a --amend -v  # squash the old HTML into your commit
  $ git push --force $YOUR_PUBLIC_REMOTE line-endings

The last push is forced, so it's “update the remote's line-endings
branch to match my local copy.  I don't care if that loses remote
history”, which we need to replace your remote commit with the new
amended version.

iglpdc:
Exactly ;-) If you could drop the html, we can merge it.

daisieh:
Okay that is fancy Git voodoo. I will try.

daisieh:
I strongly disapprove of this `HEAD^` rigmarole.

iglpdc:
Great @daisieh! I'm merging.

PR #95, opened 2015-04-20
Word/character replacement and addition of code markup

konrad:


02-create.md
```diff
@@ -17,7 +17,7 @@ $ mkdir planets
 $ cd planets
 ~~~
 
-and tell Git to make it a [repository](reference.html#repository)&mdash;a place where
+and tell Git to make it a [repository](reference.html#repository) &ndash; a place where
```
daisieh: Actually, neither of these is correct. Can you make it into a comma?

gvwilson: Why isn't the em-dash correct?

konrad: My mistake - em-dash seems to be correct (http://www.thepunctuationguide.com/em-dash.html).

daisieh: It's not that you *couldn't* use an em-dash there, it just seems unnecessary. A comma would suffice and be accurate. There is no need for emphasis of the appositive phrase.

daisieh: But it's up to you: either revert this back to the em-dash or use a comma, please.

iglpdc:
Thanks @konrad, I like it. However, I'm not sure if adding the configuration for emacs and vim is useful here. We're trying to give instructions only for the editors we advice to install for our workshops (the list is already too long for me.) Furthermore, emacs and vim are advanced editors that our learners don't typically use.

konrad:
Thanks for the feedback @iglpdc. I thought that people who might be familiar with Emacs or Vim might be not necessarily the audience of the course. Stil I gave it a try. I will remove them.

iglpdc:
Thanks, @konrad. I'm merging it.

konrad:
Great! Many thanks, @iglpdc.

PR #98, opened 2015-04-21
Updating 'Creating a Repository'

weaverbel:
Making explicit the need to move to the new directory in order to initialise it.

Issue #96, opened 2015-04-21
Commits from #66 seem lost


iglpdc:
I'll take a lot tomorrow, but I think we accidentally removed/reverted some merged commits.

daisieh:
The 01-backup edit got lost in the split of the files: they need to be placed into 02-create.
The instructors edit got lost in this commit: https://github.com/swcarpentry/git-novice/commit/eb5160d3f325d7ed1f5e679f14f6b123eb3aa12b#diff-f93309feecc5f30e1e60ad085e650818

iglpdc:
Fixed by #125.

PR #100, opened 2015-04-22
Adding change directory explanation

weaverbel:
Making explicit the need to change directory before initialising

### DIFF
--- a/02-create.md
+++ b/02-create.md
@@ -32,7 +32,7 @@ $ ls
 ~~~
 
 But if we add the `-a` flag to show everything,
-we can see that Git has created a hidden directory within planets called `.git`:
+we can see that Git has created a hidden directory within `planets` called `.git`:
 
 ~~~ {.bash}
 $ ls -a
@@ -73,4 +73,4 @@ nothing to commit (create/copy files and use "git add" to track)
 > git init     # make the beta sub-directory a Git repository
 > ~~~
 > 
-While it is possible to create one git repository inside another - there is nothing to stop your doing that - this is not a good idea. Can you explain why? 
+While it is possible to create one git repository inside another--- there is nothing to stop your doing that --- this is not a good idea. Can you explain why? 
--- a/02-create.md
+++ b/02-create.md
@@ -73,4 +73,4 @@ nothing to commit (create/copy files and use "git add" to track)
 > git init     # make the beta sub-directory a Git repository
 > ~~~
 > 
-While it is possible to create one git repository inside another--- there is nothing to stop your doing that --- this is not a good idea. Can you explain why? 
+While it is possible to create one git repository inside another&mdash;there is nothing to stop your doing that&mdash;this is not a good idea. Can you explain why? 
--- a/02-create.md
+++ b/02-create.md
@@ -10,14 +10,14 @@ minutes: 10
 
 Once Git is configured,
 we can start using it.
-Let's create a directory for our work:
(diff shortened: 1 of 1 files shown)


iglpdc:
Hi, @weaverbel , thanks for your PR. I like it, but we have to make a couple of changes before merging it.

First of all, we only change the markdown files and not the html files. This is because Github generates the html automatically. In our case, it's slightly more complicated, because we have some extra customizations, so lesson maintainers are the ones that generate the html. The bottom line is that when you do a PR you will never have to care about the html: just modify the markdown files. Because of that, I'm going to close your other PR (#98) without merging.

Second, and this is our fault, the file you are changing is slightly outdated. The challenge at the end of the file shouldn't be there: there's a newer version, which got lost, but we are going to put back in (see #96). You can leave your changes here, but there will be overwritten once we put the lost part in.

One thing that it's nice to do is to make smaller commits. In your case, for example, you have one commit, but are doing two different changes: explaining the change of directory and rephrasing the last challenge. These should be in two separate commits. The reason for that is that maybe we want to apply one and not the other and also having commit messages that are more descriptive. Don't worry about this now, but keep it in mind for future PRs.

The last thing is that we use monospaced fonts for code and names of files and folders, so your planets should be `planets` (putting the word between backticks).

So, if you can make this last change, we'll merge it. You don't need to make a new PR for the change, just make a new commit to your `weaverbel-patch-2` branch.

weaverbel:
Thanks Ivan
All good points and useful to know. I’ll try to get on to it tonight.

regards
Belinda

Belinda Weaver
eResearch Analyst Team Leader
Queensland Cyber Infrastructure Foundation
The University of Queensland
Brisbane AUSTRALIA 4072
e: b.weaver@qcif.edu.au<mailto:b.weaver@uq.edu.au> | p: +617 334 64223 | m: 0408 841 882 | w: www.qcif.edu.au<http://www.qcif.edu.au/> | Twitter @cloudaus<https://twitter.com/cloudaus>
I am based in the Research Computing Centre, level 5, Axon Building (Bldg 47)

weaverbel:
OK, Ivan I think I have made the change you want. Sorry to have caused so much bother.

regards
Belinda

________________________________

02-create.md
```diff
@@ -32,7 +32,7 @@ $ ls
 ~~~
 
 But if we add the `-a` flag to show everything,
-we can see that Git has created a hidden directory called `.git`:
+we can see that Git has created a hidden directory within planets called `.git`:
```
daisieh: This `planets` should also be in backticks.

02-create.md
```diff
@@ -73,4 +73,4 @@ nothing to commit (create/copy files and use "git add" to track)
 > git init     # make the beta sub-directory a Git repository
 > ~~~
 > 
-> Why is it a bad idea to do this?
+While it is possible to create one git repository inside another - there is nothing to stop your doing that - this is not a good idea. Can you explain why?
```
daisieh: Can you change these hyphens to em-dashes?

weaverbel:
I have now added the extra tick and changed to em dashes.

02-create.md
weaverbel: Fixed
Belinda

02-create.md
weaverbel: Fixed
Belinda

iglpdc:
Thanks, @weaverbel. All looks good now: I'm merging it.

PR #97, opened 2015-04-21
New images

daisieh:
This should address issues #87 and #85, as well as unify the images under a single visual metaphor.

PR #101, opened 2015-04-24
resize git_staging.svg to render better

msarahan:
This graphic was drawing itself way too large on my PC, and going way outside of the box.  Resizing it here hopefully hasn't ruined how it appears on any other resolution displays (but please reject this PR if it does!)


r-gaia-cs:
@msarahan Thanks for this pull request. I have a pull request to change the CSS, https://github.com/swcarpentry/styles/pull/28, that will solve this problem.

@daisieh and @iglpdc I will merge some some changes on the upstream CSS and lesson template by Sunday.

msarahan:
Sounds like a nicer solution. Closing.

daisieh:
It is also resized in #97.

PR #94, opened 2015-04-20
Working on the "explain why ignore" objective within this lesson

wcornwell:

PR #103, opened 2015-04-25
Add detail on why files should be ignored

msarahan:
This expands the intro section a bit to describe more on what kinds of files should be ignored, and why.

Also, cosmetic change to code: indent several sections to 80 character wrap


msarahan:
Closing because #94 is pretty much the same thing.

PR #104, opened 2015-04-30
change first sentence to reference the goal for the section

jjborrelli:
The previous sentence seemed like it was missing a phrase, so I added one that I thought made it more clear. 

I am issuing this pull request as part of instructor training, what do you think?


daisieh:
Seems straightforward enough. Thanks!

PR #105, opened 2015-05-02
2015-05-01 SWC at SFU, python room

daisieh:
This is what I taught for the git workshop in the Python parallel session. I probably went a little too fast; I should have done more repetitions of owner/collaborator push/pulls in the roleplay, but overall, I think it went pretty well! The new graphics seemed to be successful as visual aids.


daisieh:
This is for archival purposes only; closing.

PR #109, opened 2015-05-06
Use snapshot analogy to clarify 'staging area'

jni:
Someone asked me a question about the "staging area" vs "git commit -a" at a recent workshop, and I came up with the analogy of someone taking pictures for posterity. Then, "git add" gets the right people sitting in front of the stage for taking the picture, "git commit" actually takes the picture, and "git commit -a" says "Ok, now one with everyone!"

I think it worked pretty well. It doesn't look like the lesson covers "git commit -a" but I thought the analogy might be useful anyway to explain the staging area by itself.

### DIFF
--- a/04-changes.md
+++ b/04-changes.md
@@ -280,13 +280,14 @@ but not yet committed.
 > `git add` specifies *what* will go in a snapshot
 > (putting things in the staging area),
 > and `git commit` then *actually takes* the snapshot, and
-> copies it to long-term storage (as a commit).
+> makes a permanent record of it (as a commit).
 > If you don't have anything staged when you type `git commit`,
 > Git will prompt you to use `git commit -a` or `git commit --all`,
 > which is kind of like gathering *everyone* for the picture!
-> However, it's often considered bad practice to do this instead
-> of explicitly adding things to the staging area, because you might
-> commit changes you forgot you made. Try to stage things manually,
+> However, it's almost always better to
+> explicitly add things to the staging area, because you might
+> commit changes you forgot you made.
+> Try to stage things manually,
 > or you might find yourself searching for "git undo commit" more
 > than you would like!
 
--- a/04-changes.md
+++ b/04-changes.md
@@ -275,14 +275,14 @@ the current [change set](reference.html#change-set)
 but not yet committed.
 
 > ## Staging area {.callout}
-> If you think of git as taking snapshots of changes over the life of a
+> If you think of Git as taking snapshots of changes over the life of a
 > project,
 > `git add` specifies *what* will go in a snapshot
 > (putting things in the staging area),
 > and `git commit` then *actually takes* the snapshot, and
 > copies it to long-term storage (as a commit).
 > If you don't have anything staged when you type `git commit`,
-> git will prompt you to use `git commit -a` or `git commit --all`,
+> Git will prompt you to use `git commit -a` or `git commit --all`,
 > which is kind of like gathering *everyone* for the picture!
 > However, it's often considered bad practice to do this instead
 > of explicitly adding things to the staging area, because you might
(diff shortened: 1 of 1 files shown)


iglpdc:
Thanks, @jni! Actually, this was discussed a couple of times already. It seems that some instructors have an analogy along these lines. Mine is that you want to take a picture of the whole class. Instead of taking one shot of the whole group, risking that someone ain't looking, has his eyes shut, etc... you take individual pictures and then put them together as in a school year-book.

Changes in one file and `git add` the file would be making a pretty face and taking the individual pic, and `git commit` would be making the year-book. I don't like to explain `git commit -a` to novices: it's a shortcut that bypasses the index, which I think is a good feature in `git`. Older VCS like SVN have only `svn commit` and when you have a complex project with several files changing at once, it's hard to be sure that the commit has everything you want to.

What about expanding your PR to add a callout box with the picture analogy? 

I know that @jiffyclub has a similar (maybe the same) analogy. Can others add their tricks here so we don't miss a good explanation? I think this was long needed.

jni:
Sure, I'll wait for a few more comments and make it a callout. (That was my original idea but I hadn't seen any in this page. I see a few in other pages now, so I can model it after those.) Thanks for the support, @iglpdc!

I'll mention "git commit -a", because it's present in a few of the so-called help messages from git. ;) So students often ask about it even if you don't tell them. But I will mention that it's good practice to manually add your files.

iglpdc:
Ok, great! If you think that you can have it Monday, it'd be great. :-) We're releasing version 5.3 of the lessons and it would be nice if this is part of it. If you're busy, no worries, we add it into the next one.

jni:
Done! -ish! I'm kind of not fully satisfied with the wording, but I figure it can be refined in the future... Let me know if you have additional suggestions!

04-changes.md
```diff
@@ -269,12 +269,26 @@ but *not* commit the work we're doing on the conclusion
...
-Git has a special staging area
+Git has a special *staging area*
 where it keeps track of things that have been added to
 the current [change set](reference.html#change-set)
 but not yet committed.
-`git add` puts things in this area,
-and `git commit` then copies them to long-term storage (as a commit):
+
+> ## Staging area {.callout}
+> If you think of git as taking snapshots of changes over the life of a
```
iglpdc: git -> Git

04-changes.md
```diff
@@ -269,12 +269,26 @@ but *not* commit the work we're doing on the conclusion
...
+> project,
+> `git add` specifies *what* will go in a snapshot
+> (putting things in the staging area),
+> and `git commit` then *actually takes* the snapshot, and
+> copies it to long-term storage (as a commit).
```
iglpdc: Maybe "copies it to long-term storage" -> "makes a permanent record of the changes"

04-changes.md
```diff
@@ -269,12 +269,26 @@ but *not* commit the work we're doing on the conclusion
...
+> If you don't have anything staged when you type `git commit`,
+> git will prompt you to use `git commit -a` or `git commit --all`,
+> which is kind of like gathering *everyone* for the picture!
+> However, it's often considered bad practice to do this instead
```
iglpdc: Maybe "it's often considered bad practice to do this instead of explicitly adding things to the staging area" -> "it's always better to explicitly add things to the staging area"

iglpdc:
Hey, @jni, thanks for the new commits. I made a couple of comments, mostly minor rewording. The other thing is that we use "Git" instead of "git" when is not a command.

jni:
Thanks for the comments @iglpdc! I've also added a little humorous sentence to illustrate why "-a" is bad, but I'm happy to remove that commit if you don't think it fits! Let me know!

iglpdc:
I like it! Merging.

PR #115, opened 2015-05-08
Add missing sub-headers for the last two topics.

iglpdc:
There were a couple of missing subheadings.


iglpdc:
I'm actually merging this because it's trivial.

PR #116, opened 2015-05-08
Split 06-collab in two different topics

iglpdc:
This is a fix to #113, if we finally decide to do it.

I've chosen "Remotes In Github" as the name for the new topic. Other suggestions are welcome.

The rest is just renaming the files, and moving things around.

PR #124, opened 2015-05-09
Split "Setting up" in two topics

iglpdc:
"Setting up" contains two different parts: first, a motivation and basics of a VCS, and, second, instructions to set up Git locally. This splits the file in two.

It also fixes a warning about having subheading on the original file.

Feedback on whether we want to make the split or not are welcome.

**Don't merge yet** as this is dependent on #116.


02-setup.md
```diff
@@ -9,38 +9,6 @@ minutes: 10
 > *   Explain which initialization and configuration steps are required once per computer,
```
iglpdc: Why are we saying "configuration per computer", and "per repository". We only explain here the steps needed to configure Git "per computer". If this is refer to the `--global` file, I think the learning objectives can be simplified to:

- Configure `git` the first time is used on a computer
- Understand the meaning of the `--global` configuration flag

BernhardKonrad:
This is great! +1 for merging.

msarahan:
+1 looks good to me.

PBarmby:
+1 as well.

iglpdc:
Thanks @BernhardKonrad and @msarahan. I'm merging then.

swaldman3:
Agree with the split - especially useful for people after a course who want a quick reminder of the commands to configure git but don't need to go through the motivations again.

PR #125, opened 2015-05-09
Add back the changes in c41ed8c0d

iglpdc:
They were merged in #66, but lost at some point (see #96). The other commit
of #66 was not lost.


iglpdc:
I'm merging this because is a change that was already merged.

PR #126, opened 2015-05-09
Fix broken links: links to .md -> .html files

iglpdc:


iglpdc:
I'm merging because it's trivial.

PR #114, opened 2015-05-08
Reformat, reorder, and add new tips to the teaching guide.

iglpdc:
This fixes #112.

PR #127, opened 2015-05-10
Updated html so that index doesn't have dead links.

jarthurgross:
Ran `make preview` to generate the html from the md files so that the main page isn't full of dead links like it currently is. Also fixed broken reference to renumbered link to history in instructors.md.

### DIFF
--- a/01-basics.html
+++ b/01-basics.html
@@ -1,0 +1,81 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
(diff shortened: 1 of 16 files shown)


iglpdc:
Thanks, @jarthurgross. 

Maintainers update the html, so it's easier to avoid conflicts (we'll do it today or tomorrow, after merging all the PRs for version 5.3). The sentence with the broken link was removed in PR #114 (that I just merged), because it was a bit redundant, given that now the guide is split into sections.

I'm closing without merging.

Issue #111, opened 2015-05-06
Show how to delete and move files


KathleenLabrie:
This tutorial almost perfect for my target audience.  The only thing missing, I think, is to show how to properly delete and move files around.  Should be fairly straightforward to add something like that to the "change" section.

r-gaia-cs:
@KathleenLabrie Thanks for the report.

> The only thing missing, I think, is to show how to properly delete and move files around.

Git is very smart and identify deleted and moved files:

~~~
$ git init
Initialized empty Git repository in /tmp/sample/.git/
$ touch foo.txt
$ git add foo.txt 
$ git commit -m 'Create foo.txt'
[master (root-commit) e6d02a3] Create foo.txt
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 foo.txt
$ mv foo.txt bar.txt
$ git status
On branch master
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        deleted:    foo.txt

Untracked files:
  (use "git add <file>..." to include in what will be committed)

        bar.txt

no changes added to commit (use "git add" and/or "git commit -a")
$ git add foo.txt bar.txt 
$ g status
On branch master
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

        renamed:    foo.txt -> bar.txt
~~~

> Should be fairly straightforward to add something like that to the "change" section.

Can you send a pull request with the challenge? If yes and you need help, let us know.

gvwilson:
We should introduce `git mv` and `git rm`, and also say that Git can 
figure out moves and deletes.

iglpdc:
I'm -1 to add this to the novice lesson, although I agree it's valuable. I think that moving and deleting files is not something novices will face the first few times they use a repo, and the lesson it's pretty hard to teach in 3 hours already.

If we want to include move and delete, we will need to explain `git rm` and `git mv` (which involves at least 3 commits, see the amount of code in @r-gaia-cs 's example above), and why/how Git tracks the file when we use just `mv` (or maybe a bit on how Git keeps track of changes). I think this will add another 5-10 minutes to the lesson and will probably require another topic.

This could be good for the intermediate lesson or extras in this topic, depending on how deep we want to go on explaining how Git does the things under hood.

KathleenLabrie:
[...]

PR #146, opened 2015-05-16
Section on removing and renaming files (Issue #111)

KathleenLabrie:
As discussed in Issue #111, I added two sections on removing files and renaming files to the Discussion page.


discussion.md
```diff
@@ -201,3 +201,145 @@ Even though a logo will be stored in a binary format such as `jpg` or `png`,
...
+~~~
+~~~{.output}
+On branch master
+nothing to commit, working directory clean
+~~~
+
+Nibiru is not a real planet.  That was a silly idea.  Let us remove
+it from the disk and then from the repository:
+
+~~~{.bash}
```
iglpdc: You don't need to do this. `git rm` removes the file from the working tree  and the index [1], meaning that it does a `rm nibiru.txt` when you do a `git rm nibiru.txt` 

[1] http://git-scm.com/docs/git-rm

discussion.md
```diff
@@ -201,3 +201,145 @@ Even though a logo will be stored in a binary format such as `jpg` or `png`,
...
+Changes not staged for commit:
+   (use "git add/rm <file>..." to update what will be committed)
+   (use "git checkout -- <file>..." to discard changes in working directory)
+
+    deleted:    nibiru.txt
+
+no changes added to commit (use "git add" and/or "git commit -a")
+~~~
+
+See how git has noticed that the file `nibiru.txt` has been removed
```
iglpdc: We use capitalized "Git" unless is the command itself, so "git" ->  "Git".

discussion.md
```diff
@@ -201,3 +201,145 @@ Even though a logo will be stored in a binary format such as `jpg` or `png`,
...
+Changes to be committed:
+   (use "git reset HEAD <file>..." to unstage)
+
+   deleted:    nibiru.txt
+
+~~~
+
+The change has been staged.  Now commit the removal, and remove the
+file from the repository itself.  Note that the file will be removed
+in the version you are committing.  The previous revision will still
```
iglpdc: Maybe "from" instead of "in" to mean that it will be deleted in this version and all the ones after.

Also, "The previous revision" -> "The previous commit" (we prefer to use commit if possible).

discussion.md
```diff
@@ -201,3 +201,145 @@ Even though a logo will be stored in a binary format such as `jpg` or `png`,
...
+have the file, if you were to retrieve that specific revision.
```
iglpdc: Again "revision" -> "commit"

discussion.md
```diff
@@ -201,3 +201,145 @@ Even though a logo will be stored in a binary format such as `jpg` or `png`,
...
+
+Untracked files:
+  (use "git add <file>..." to include in what will be committed)
+
+    earth.txt
+
+no changes added to commit (use "git add" and/or "git commit -a")
+~~~
+
+`git` has noticed that the file `krypton.txt` has disappeared from the
```
iglpdc: "`git`"-> "Git"

discussion.md
```diff
@@ -201,3 +201,145 @@ Even though a logo will be stored in a binary format such as `jpg` or `png`,
...
+~~~{.output}
+On branch master
+Changes to be committed:
+  (use "git reset HEAD <file>..." to unstage)
+
+    renamed:    krypton.txt -> earth.txt
+
+~~~
+
+Notice how `git` has now figure out that the `krypton.txt` has not
```
iglpdc: "`git`" -> "Git"

discussion.md
```diff
@@ -201,3 +201,145 @@ Even though a logo will be stored in a binary format such as `jpg` or `png`,
...
+
+~~~{.bash}
+$ git add krypton.txt
+$ git commit -m 'Adding planet Krypton'
+~~~
+
+We all know that Superman moved to Earth.  Not that he had much
+choice.  Now his home planet is Earth.
+
+Rename the file `krypton.txt` to `earth.txt`:
```
iglpdc: Why not use `git mv krypton.txt earth.txt`? It works the same as `git rm`, first does the change in the working directory `mv krypton.txt earth.txt` and then in the index. [1]

[1] http://git-scm.com/docs/git-mv

discussion.md
```diff
@@ -201,3 +201,145 @@ Even though a logo will be stored in a binary format such as `jpg` or `png`,
...
+
+~~~{.bash}
+$ mv krypton.txt earth.txt
+~~~
+
+Let us see what `git` has figured out by itself:
```
iglpdc: "`git`" -> "Git"

iglpdc:
Thanks @KathleenLabrie ! It's a good addition, as we talked before. I made some comments. A couple of style things: we use capitalized "Git" and prefer "commit" to version and revision. 

The other thing is that you can save the Unix `mv` and `rm`, is there any reason why you want to do it otherwise?

KathleenLabrie:
Noted about "Git" and "commit".  This is my first pull request ever, have you made the changes already or have you left them to me?  (I'm more than happy to do it, of course, that's just fair.)

Regarding the use of Unix `mv` and `rm`, I used them with my target audience in mind.  New Git users already very familiar with Unix will 99% of the time move and delete with the Unix command.  They will forget to use `git mv` and `git rm`.  Git is seen as the thing you use not "to make changes" but "to record changes".    That's one thing I like a lot about Git versus SVN.  In SVN, you have no choice and when the Unix command are used instead of the SVN equivalent, you are in a mess.  With Git, no problem, it figures it out.  Both ways work.

Also, when we create a file, it is first created with a Unix command and only then do we record the change with Git.  So doing the move and delete the Unix way then record, does mirror the creation and modification of a file.

Anyway, that's my point of view based on my experience and what I know of my target audience.  To me, it seems like a more natural workflow.  We can change the example to use only Git if it is preferred.

KathleenLabrie:
Hi @iglpdc, I have finally found time to fix the text.  SciPy conferences are good for that.  I have fixed the Git and revision/version style issues.   I have also made the 'git rm' and 'git mv' form the "default".  Then I have put the Unix form in a subsection and presented as a "if you forgot to use git, here's how your recover".   I hope this addresses your valid concerns.

iglpdc:
I like it. I'm merging. Thanks, @KathleenLabrie.

PR #151, opened 2015-06-06
Update link to instructions

gvwilson:
Pointing at lesson-example instead of lesson-template, and telling people (again) not to commit HTML.


iglpdc:
Thanks, @gvwilson!

PR #154, opened 2015-06-11
Modification to sentence starting with "As a bonus..." (made to markdown file)"

Haschmi:
Closed the old request and opening a new one where the markdown file was changed.
Sorry for the confusion, new to both GIT and markdown files.

### DIFF
--- a/06-ignore.html
+++ b/06-ignore.html
@@ -74,7 +74,7 @@ $ <span class="kw">git</span> commit -m <span class="st">&quot;Add the ignore fi
 $ <span class="kw">git</span> status</code></pre>
 <pre class="output"><code># On branch master
 nothing to commit, working directory clean</code></pre>
-<p>As a bonus, using <code>.gitignore</code> helps us avoid accidentally adding files to the repository that we don’t want to track:</p>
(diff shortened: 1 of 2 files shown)


gvwilson:
:+1: from me - thanks!

iglpdc:
Hi @Haschmi. Thanks for catching this. I have a suggestion, though. Don't you think it would be better: "adding files that we don't want to track to the repository" -> "adding to the repository files that we don't want to track". Also you could drop the "that" in the second version.

If you think it sound better, you can make a new commit to your local repo and the push it your remote in Github (you don't need to make a new pull request: it would be automatically updated). It'd be something like:

(... make changes ...)
$ git add 06-ignore.md
$ git commit
$ git push origin gh-pages

Then you can head to your pull request (this very same page) and see the updates.

Haschmi:
Hi Ivan;

I agree, this is the better way of putting it. The sentence I proposed has got too much German in it. I made the change and pushed it.

Regards, Hartmut

iglpdc:
Great! Thanks again for your work. I'm merging.

PR #157, opened 2015-06-23
add clearer instructions to setup remote for a new repo

maneesha:
The instructions as they are apply just to existing repos.  The item to click in the  image in Figure: Changing the Repository URL on GitHub does not show for a newly created, empty repo.  I added notes to follow the instructions on the new repo's set up page.  I'm inclined to remove the instructions that assume it's an existing repo entirely since we are starting from a new repo.  What do others think?

### DIFF
--- a/07-github.md
+++ b/07-github.md
@@ -58,7 +58,7 @@ but the remote repository on GitHub doesn't contain any files yet:
 The next step is to connect the two repositories.
 We do this by making the GitHub repository a [remote](reference.html#remote)
 for the local repository.
-The home page of the repository on GitHub includes
+For existing repositories, the home page of the repository on GitHub includes
 the string we need to identify it:
 
 ![Where to Find Repository URL on GitHub](fig/github-find-repo-string.png)
@@ -79,7 +79,7 @@ Click on the 'HTTPS' link to change the [protocol](reference.html#protocol) from
 
 Copy that URL from the browser,
 go into the local `planets` repository,
-and run this command:
+and run this command. For a newly created repository, the GitHub setup page should give you the same command.
 
 ~~~ {.bash}
 $ git remote add origin https://github.com/vlad/planets
@@ -104,7 +104,7 @@ but `origin` is by far the most common choice.
 
 Once the nickname `origin` is set up,
 this command will push the changes from our local repository
-to the repository on GitHub:
+to the repository on GitHub.  Again, for a newly created repository, you'll see this command on the set up page.
 
 ~~~ {.bash}
 $ git push origin master


07-github.md
```diff
@@ -79,7 +79,7 @@ Click on the 'HTTPS' link to change the [protocol](reference.html#protocol) from
 
 Copy that URL from the browser,
 go into the local `planets` repository,
-and run this command:
+and run this command. For a newly created repository, the GitHub setup page should give you the same command.
```
r-gaia-cs: Can we move the new sentence for after the command?

07-github.md
```diff
@@ -104,7 +104,7 @@ but `origin` is by far the most common choice.
 
 Once the nickname `origin` is set up,
 this command will push the changes from our local repository
-to the repository on GitHub:
+to the repository on GitHub.  Again, for a newly created repository, you'll see this command on the set up page.
```
r-gaia-cs: Same here. Can we move the new sentence for after the command?

iglpdc:
@maneesha, thanks for catching this. I agree the current version is confusing.

But, I think the change can be confusing too. We ask the learners to create the repo on Github at the beginning of these topic, so the repo in Github exists. The Github instructions (new versus existing repo) refer to an existing **local** repo. This is the case we are dealing with, as we have the local `planets` in our machine previously from the lesson. 

The current change seems to imply that the "existing" and "new repo" refer to the Github repo, which is not true. At least this is what I understood.

I agree though that the explanation of where to find the clone URL could be improved and the whole thing is not in order. So, what about:

- moving the text starting by "As soon as ..." and the second screenshot after the "Freshly-made Github repository", and
- moving the screenshots with the clone URL down, right after the "Github repository after the first push", 

In this way, we have the whole thing ordered and at the end we have a separate piece indicating how to get the clone URL for any repo.

maneesha:
​Looking at this more closely, ​
I think all this actually needs a lot more work...
​One thing - ​
The screenshot for Figure: Creating a Repository on GitHub (Step 1)
also needs to be updated as GitHub has changed its UI since this was
created.

For the rest of it,  if the lesson is followed closely, students will have
an existing local repo which they'll be connecting to a new remote repo so
I think the focus should be on how to do this.  We could have additional
instructions as a reference for how to set up a remote for an existing
local repo to an existing remote but it is confusing to have it all here as
it is now.

​I also think, for the benefit of new users, we should also include an
explanation for the other options you see when you set up a new repo on
Github (public vs private, initialize with a Readme, etc.)

​I can work on revamping all of this over the next few days and share a new
pull request.

Any other thoughts, let me know.

​


​





On Thu, Jun 25, 2015 at 10:07 PM, Ivan Gonzalez <notifications@github.com>
wrote:
>
> @maneesha, thanks for catching this. I agree the current version is
confusing.
>
> But, I think the change can be confusing too. We ask the learners to
create the repo on Github at the beginning of these topic, so the repo in
Github exists. The Github instructions (new versus existing repo) refer to
an existing local repo. This is the case we are dealing with, as we have
the local planets in our machine previously from the lesson.
>
> The current change seems to imply that the "existing" and "new repo"
refer to the Github repo, which is not true. At least this is what I
understood.
>
> I agree though that the explanation of where to find the clone URL could
be improved and the whole thing is not in order. So, what about:
>
> moving the text starting by "As soon as ..." and the second screenshot
after the "Freshly-made Github repository", and
> moving the screenshots with the clone URL down, right after the "Github
repository after the first push",
>
> In this way, we have the whole thing ordered and at the end we have a
separate piece indicating how to get the clone URL for any repo.

daisieh:
I'm closing this, since @maneesha will make a new PR for the updated version.

PR #160, opened 2015-06-26
Revise assignment in part 7. In particular, eliminate the word "clone".

arokem:
As far as I could tell, this is the first time this command appears in this
lesson, and it seems that it should be explained before it assigned in an
exercise.

### DIFF
--- a/07-github.md
+++ b/07-github.md
@@ -182,9 +182,8 @@ this command would download them to our local repository.
 
 > ## GitHub Timestamp {.challenge}
 >
-> Create a repository on GitHub,
-> clone it,
-> add a file,
-> push those changes to GitHub,
-> and then look at the [timestamp](reference.html#timestamp) of the change on GitHub.
+> Create a remote repository on GitHub.
+> Push the contents of your local repository to the remote.
+> Make changes to your local repository and push these changes.
+> Look at the [timestamp](reference.html#timestamp) of the change on GitHub.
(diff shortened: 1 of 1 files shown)


arokem:
Besides, what is the answer to the question asked at the end of this exercise?

iglpdc:
Hi, @arokem, thanks for the PR. I agree that `clone` shouldn't be introduced here. I guess this was at the end of the lesson before we split it.

The point of the exercise, as I see it, is to stress the idea that the repo in Github is a copy of your repo, history included, and not just a  copy of your files. So, the time stamps in Github reflect the time when you created the repo locally (even if this was months ago.) This may be counter intuitive for people: when you copy a file somewhere usually the timestamps reflect the time of the copy.

As for your changes, what do you (and the others) think about simplifying even more and say something like "go to the repo you just created in Github and check the timestamps"?

arokem:
OK - I understand the question/answer now, but the point is lost here, because the repo and the files were both just created, so there is no distinction. How about changing things up a bit. For example, how about recycling question 2 from here? 

https://github.com/swcarpentry/recycling/blob/gh-pages/_posts/2014-10-01-assessment-collaborating-with-git.md

For the time being, I liked your suggestion for wording and implemented that.

iglpdc:
Thanks @arokem! (and sorry for the delay in merging)

PR #165, opened 2015-07-09
More precise statement.

lduchesne:
This might not be the best way to address it, but I think the original "settings for every project on this computer" is a bit misleading, as that statement would apply to the --system flag.


iglpdc:
Yes, you're right. The problem is that some of our learners don't have a clear picture of the system vs user settings. In any case, I think your change is clarifying. I'm merging it.
Thanks, @lduchesne !

PR #166, opened 2015-07-10
metadada to metadata

marwahaha:


iglpdc:
Oh, good catch! Thanks @marwahaha.

PR #172, opened 2015-09-02
Quality Control Fixes and Title Changes

preney:
After cloning the git-novice repository, "make check" returned errors, i.e.,

    python tools/check.py .
    ERROR: In ./discussion.md: Heading at line 254 should be level 2
    ERROR: In ./discussion.md: Heading at line 350 should be level 2
    ERROR: In ./discussion.md: Found callout box with unrecognized style 'callout'
    WARNING: 3 errors were encountered during validation. See log for details.
    Makefile:38: recipe for target 'check' failed
    make: *** [check] Error 3

The first two were fixed and the third appears to be a bug with "make check" itself so it was left as-is. The title for the callout box was edited to be correct for its content and the two other headings were edited to clearly say files are being manipulated.

### DIFF
--- a/discussion.html
+++ b/discussion.html
@@ -63,7 +63,7 @@ $ <span class="kw">git</span> config --global --unset log.abbrevCommit
 $ <span class="kw">git</span> config --global --unset format.pretty</code></pre>
 <aside class="callout panel panel-info">
 <div class="panel-heading">
-<h2><span class="glyphicon glyphicon-pushpin"></span>Undoing Git Configuration Changes</h2>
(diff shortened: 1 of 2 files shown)


iglpdc:
Thanks, @preney. Sorry for the delay in taking care of this. Could you please drop the commit that changes the html file? We generate the html in a different way and you don't need to include it in your commit.

preney:
@iglpdc I reverted the discussion.html file back to what it was. Is this sufficient for what you want/need?

iglpdc:
Yes, @preney. Thanks for your contribution, I'm merging.

PR #177, opened 2015-09-30
https git urls end with .git

jiffyclub:
screenshot 2015-09-30 15 00 40


iglpdc:
Thanks @jiffyclub!

drj11:
For information rather than action:

I suppose it's better pedagogy to show what github show, but in fact it works with or without the `.git` and after a while I got used to typing them rather than pasting them, then it's nice not to have to use the `.git`.

Issue #162, opened 2015-07-02
git in RStudio


noamross:
In workshops teaching R through RStudio, it's becoming a more common pattern to teach this git command line lesson, then follow it with a short lesson demonstrating the git GUI interface in RStudio.  It would be useful to have this "transition" lesson written up with screenshots.  While RStudio isn't quite a full-featured git GUI it is the environment many students will use.

This could also live in the r-novice repositories.

Here are some lessons designed for teaching git only through RStudio which could be adapted:

1. https://github.com/microbialinformatics/assignment01
2. https://jennybc.github.io/2014-05-12-ubc/ubc-r/session03_git.html

There are a couple of technical issues for this:

1.  RStudio may need to be pointed to the git executable, especially on Windows machines. Instructions: https://stat545-ubc.github.io/git03_rstudio-meet-git.html 
2.  Remote tracking must be set up on the repository using the `-u` flag, e.g. `git push -u origin master`.
3.  If students want github passwords to be stored, this will need to be activated.  This may not be necessary during the workshop but one could point them to a resource showing how for later: https://stat545-ubc.github.io/git06_credential-caching.html

iglpdc:
I agree it could be a good thing to have some extra topics on how to use git from the interfaces provided by other tools. These materials e.g. "Using Git with Rstudio" or "Using Git with Matlab" (which has also a similar GUI for Git), could be linked in the main lesson, but not be part of it, so instructors and solo learners can follow the link if the are using RStudio or Matlab.

iglpdc:
I'm closing this after merging #183.

PR #183, opened 2015-10-30
Added section on RStudio integration

emilydolson:
As suggested in https://github.com/swcarpentry/git-novice/issues/162, here's a first pass at a supplemental section on using Git with RStudio. It continues with the planets repository example, and has learners create a project in RStudio from their planets directory, make a change, commit it, and push it.

Some things I'm not sure about:
* File naming convention - since this is a supplemental lesson, what should the markdown file be called?
* Handling potential configuration issues - right now I just have callouts pointing to resources for resolving potential problems. I considered including a more detailed description of finding the git executable in the callout, or even integrating it into the main text (i.e. have learners start by making sure it's configured right), but there's a trade-off between providing more information and making the lesson too long. When I taught this before (same content, not technically an SWC setting), it seemed like a relatively small percentage of the learners encountered these issues, so I erred on the side of brevity. If these issues are more prevalent, though, I can easily add some step-by-step configuration instructions.
* There are a lot of Git integration features in RStudio and it's probably not worth mentioning them all. Does this seem like a reasonable subset?


iglpdc:
Thanks, @emilydolson. We don't have way to handle this type of supplemental material, but I think it's really a good addition. I'm gonna merge the PR as is and then we can decide on how to link it.

Issue #133, opened 2015-05-10
Which is the difference between "revision" and "commit"?


iglpdc:
In our [Glossary](http://swcarpentry.github.io/git-novice/reference.html), we have different entries for "revision" and "commit". "Commit" is defined as a verb, and "revision" as a noun. The problem is that even in the very definition of "commit", we use it as a noun. 

This is not a big deal, but in the lesson (see e.g. [04-changes.md](https://github.com/swcarpentry/git-novice/blob/gh-pages/04-changes.md)) we use sometimes "commit" and sometimes "revision" to mean a recorded set of changes in the repo. There's also instances of "version" and "set of changes". This could be confusing for a novice, specially if reading the lesson alone. 

I would like to have just one word (which is better "commit" or "revision"?) for the recorded set of changes in the repo, and leave "version" for a snapshot of the files not necessary related to the repo ("the new version of your thesis was trashed by your advisor").

r-gaia-cs:
>  would like to have just one word (which is better "commit" or "revision"?) for the recorded set of changes in the repo, and leave "version" for a snapshot of the files not necessary related to the repo ("the new version of your thesis was trashed by your advisor").

Use "commit" for the recorded set of changes in the repo.

iglpdc:
Yes, I agree with that. I'll PR later today.

iglpdc:
Fixed by #139

PR #186, opened 2015-11-14
reference: Replace 'revision' with 'commit'

wking:
Fixing a few instances missed by #139, based on the consensus formed
in #133.  More details on these changes (and some drive-by stuff I
noticed while making them) in the commit messages.


iglpdc:
Thanks @wking!

PR #185, opened 2015-11-14
adding information about cloning to topic 7

ChristinaLK:
I created this PR as part of a demo - probably needs further discussion!  

Created in response to #176

Issue #176, opened 2015-09-22
"Remotes in GitHub" claims to cover cloning from github but does not do so.


dham:
One of the learning objectives of this lesson is "Clone a remote repository."

However the lesson does not actually cover cloning, it only covers adding a remote to an existing repo.

daisieh:
We cover this in the next lesson; we should just move this objective there.

iglpdc:
This is fixed: we mention "cloning" only on [08-collab.md](https://github.com/swcarpentry/git-novice/blob/gh-pages/08-collab.md). Closing.

PR #191, opened 2015-11-17
Refactor to forking

daisieh:
I forgot that I meant to suggest these changes to the main repository: this is basically what we taught at https://github.com/swcarpentry/git-novice/pull/168 and I thought that the revamped conflict and motivation + the new forking lesson went really well. I'd like to suggest incorporating them into the main repo.

Also, this would address https://github.com/swcarpentry/git-novice/pull/185 / https://github.com/swcarpentry/git-novice/issues/176


daisieh:
@iglpdc, what do you think about merging these changes? I would love some more feedback before merging. I'd be happy to do the conflict resolution, though.

daisieh:
I have run the lesson all the way through as presented here several times and have always made it to the end, with pretty good feedback. I think that introducing pull requests makes a nice capstone example as well, especially if you can introduce that capstone at the beginning as a thing you will learn to do in the lesson.

daisieh:
I think that using the example of Andrew (or, you know, any other SWC instructor) is more relatable than Wolfman/Dracula, but I think that there are some residual bits left from their story in the text. I never can think of a good way to integrate them, so I just use random gibberish, or more recently, comments about Mars from _The Martian_.

daisieh:
Do people generally get to 10-open, 11-licensing, and 12-hosting when they teach this lesson? I just realized that the reason I can always get to forking is because I never teach those. I also don't teach 06-ignore.

daisieh:
I'd love to integrate this into the 5.4 lesson on bug barbeque day. What do you think, @iglpdc ?

iglpdc:
@daisieh I'm closing this by now. Feel free to remake it after 5.4.

PR #194, opened 2015-11-30
Added Apache License, version 2.0 as it is often recommended when col…

annefou:
…laborating with private companies

### DIFF
--- a/reference.md
+++ b/reference.md
@@ -17,14 +17,14 @@ subtitle: Reference
 *   `git status` shows the status of a repository.
 *   Files can be stored in a project's working directory (which users see),
     the staging area (where the next commit is being built up)
-    and the local repository (where revisions are permanently recorded).
+    and the local repository (where commits are permanently recorded).
 *   `git add` puts files in the staging area.
-*   `git commit` saves the revisions in the staging area to the local repository.
+*   `git commit` saves the staged content as a new commit in the local repository.
 *   Always write a log message when committing changes.
 
 ## [Exploring History](05-history.html)
 
-*   `git diff` displays differences between revisions.
+*   `git diff` displays differences between commits.
 *   `git checkout` recovers old versions of files.
 
 ## [Ignoring Things](06-ignore.html)
@@ -80,7 +80,7 @@ change set
     [repository](#repository).
 
 commit
-:   To record the current state of a set of files (a [change set](#changeset))
+:   To record the current state of a set of files (a [change set](#change-set))
     in a [version control](#version-control) [repository](#repository). As a noun, 
     the result of commiting, i.e. a recorded change set in a repository.
     If a commit contains changes to multiple files,
@@ -125,8 +125,7 @@ resolve
     being managed by a [version control](#version-control) system.
 
 revision
-:   A recorded [change set](#change-set) of a [version control](#version-control) 
-    [repository](#repository). The same as a [commit](#commit).
+:   A synonym for [commit](#commit).
 
 SHA-1
 :   [SHA-1 hashes](http://en.wikipedia.org/wiki/SHA-1) is what Git uses to compute identifiers, including for commits.


wking:
On Mon, Nov 30, 2015 at 12:08:43AM -0800, annefou wrote:
> * Added Apache License, version 2.0 as it is often recommended when
>   collaborating with private companies

To back this up with more detail from a reliable source, I'd link to
the FSF's recommendation based on the patent termination provision
[1].

[1]: https://www.gnu.org/licenses/license-list.html#apache2

annefou:
Yes I fully agree with you. Would it be OK to add the following:

* [Apache license, Version 2.0](http://www.apache.org/licenses/LICENSE-2.0), often considered as being the best with regard to patents (see [https://www.gnu.org/licenses/license-list.html#apache2](https://www.gnu.org/licenses/license-list.html#apache2)).

wking:
On Tue, Dec 01, 2015 at 12:45:42AM -0800, annefou wrote:
> Would it be OK to add the following:

I'm a lesson maintainer, so take my perspective with a grain of salt
;).

Looking into this more closely, the cited Morin, et al. 2012 covers
Apache v2 and its patent grants with more detail than we probably want
to include here.  We don't provide further details about the MIT or
BSD licenses, so I'll reverse my initial suggestion and say:

1. We should drop cb6862f to keep the diff just an “Apache license,
   version 2.0” listing.
2. We should amend 6a3f8b9's commit message to mention the FSF's
   recommendation and Morin et al.'s coverage.

That keeps a consistent and small listing for sharing with learners,
while giving a bit more detail and support in the Git history for
*why* we felt we should mention the Apache v2.

annefou:
I am not sure entirely agree but I understand your point of view. 

Anyway, I followed your suggestion and changed the commit message. Is it clearer?

wking:
On Tue, Dec 01, 2015 at 12:36:16PM -0800, annefou wrote:
> I am not sure entirely agree but I understand your point of view.

I'm not the maintainer, so don't feel pressured to adjust your PR
based on my feedback ;).  If the maintainers have a preference,
adapting to that is good.  If they don't have a preference, do
whatever makes the most sense to you.  I'm just sharing my point of
view and reasoning to help inform your decision-making process.

> Anyway, I followed your suggestion and changed the commit
> message. Is it clearer?

You need a blank line between your commit message's summary and body
to avoid things like:

  $ git log --oneline origin/pr/194
  29866be Added Apache License, version 2.0 as it is often recommended when collaborating with private companies. For more information, see https://www.gnu.org/licenses/license-list.html#apache2 and paper written by Andrew Morin, Jennifer M. Urban, and Piotr Sliz: "A Quick Guide to Software Licensing for the Scientist-Programmer" PLoS Computational Biology 8.7 (2012).
  ce61b87 Merge pull request #186 from wking/revision-to-commit
  e772e47 reference: Trim 'revision' definition to bare minimum

More details in [1].  I'd recommend:

  licensing: Mention Apache License, version 2.0

  It is often recommended when collaborating with private
  companies.  For more information, see [1] and the Morin paper linked
  from the topic content [2].

  [1]: https://www.gnu.org/licenses/license-list.html#apache2
  [2]: http://dx.doi.org/10.1371/journal.pcbi.1002598

[1]: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

annefou:
When I said I don't fully agree it's more on whether we should have more details or not on licenses. But I agree if we add some details on Apache License 2.0, it should be done for other type of licenses. But having too much details is not the goal of this lesson... So I understand your comments (whether you are the maintainer or not!).
Thanks for the tips on commit messages (http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html). Very clear and definitely much better with the log!

11-licensing.md
```diff
@@ -47,6 +47,7 @@ the most popular, including the following:
   (GPL),
 * [MIT license](http://opensource.org/licenses/MIT),
 * [BSD license](http://opensource.org/licenses/BSD-2-Clause).
+* [Apache license, Version 2.0](http://www.apache.org/licenses/LICENSE-2.0)
```
drj11: How about linking to the opensource.org hosted version: https://opensource.org/licenses/Apache-2.0 ?

Before finding it on opensource.org, I wasn't sure if Apache license was an Open Source license or not.

annefou:
True. Apache license version 2.0 are then described/documented in a similar way.

PR #201, opened 2016-01-08
added exercise for instructor training

cloucks:


05-history.md
```diff
@@ -220,3 +220,43 @@ moving backward and forward in time becomes much easier.
...
 > 5. Both 2 & 4
+
+
+> ## Understanding Workflow and History {.challenge}
+>
+> What is the output of the following commands?
+>
+> ~~~ {.bash}
+> $ cd planets
+> $ nano venus.txt #input the following text: Venus is beautiful and full of love
```
iglpdc: What about using:

```
$ echo "Venus is beautiful and full of love" > venus.txt
```

and 

```
$ echo "Venus is too hot to be suitable as a base" >> venus.txt
````

This way we complicate a bit the exercise (difference between `>` and `>>`) and also avoid having the comment in the command line.

File: 05-history.md [hunk 2]
```diff
@@ -220,3 +220,43 @@ moving backward and forward in time becomes much easier.
 >     $ git checkout <unique ID of last commit> data_cruncher.py
 >     ~~~
 > 5. Both 2 & 4
+
+
+> ## Understanding Workflow and History {.challenge}
+>
+> What is the output of the following commands?
```
iglpdc: What about: "What is the output of `cat venus.txt` at the end of this set of commands?" Do you think that sounds a bit clearer?

05-history.md
gvwilson: Have we already introduced `echo something > file`?  If not, I think 
it'd be better to stick with editing the file.

iglpdc: We do in the shell lesson [1], but I'm also fine with the current version.

[1] http://swcarpentry.github.io/shell-novice/03-pipefilter.html

cloucks:
Thanks for the feedback @iglpdc! I modified the question based on your feedback. I like the echo idea, but I used the comments because it helped me when first learning to actually manipulate the file since I wasn't an expert in shell commands. Either way works for me though!

iglpdc:
Ok @cloucks ! Sounds reasonable then to leave the comments there. I'm merging it. Thanks again for your contribution.

PR #204, opened 2016-01-22
added author/committer challenge

timtroendle:
Added a challenge about Git's author/committer mechanism as part of the instructor training.

### DIFF
--- a/04-changes.md
+++ b/04-changes.md
@@ -478,7 +478,7 @@ repository (`git commit`):
 > $ git commit --author="Vlad Dracula <vlad@tran.sylvan.ia>"
 > ~~~
 >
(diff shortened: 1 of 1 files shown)


iglpdc:
Thanks @timtroendle! Great challenge (I actually didn't know this). I like very much that you point to use cases for this (e.g. a collaborator who refuses to use VC and still sends you changes by email).

I have one minor comment, thought. Could you make it to do just two commits, instead of three? We don't have much time during the workshops and I don't see the point of adding a second commit under your name.

timtroendle:
Happy to hear that you like it. I agree that the third commit is not necessary and removed it.

iglpdc:
Great @timtroendle ! I'm merging it.

PR #208, opened 2016-01-25
Add challenge to check understanding of push vs. commit

catherio:
Add a challenge in the github lesson to check understanding of push vs. commit.

The "07 github" lesson is the first time the "push" command is introduced. It seems like a useful moment for students to take to reflect on the difference between "push" (which they just learned) and "commit" (which they already knew), since both might just sound like "save my work" to an unfamiliar learner.


daisieh:
I think this is a good idea.

PR #209, opened 2016-01-26
Add challenge about fixing a remote with an invalid URL

lesteve:
This is part of the instructor training.

### DIFF
--- a/07-github.md
+++ b/07-github.md
@@ -192,3 +192,19 @@ this command would download them to our local repository.
 >
 > In this lesson, we introduced the "git push" command.
 > How is "git push" different from "git commit"?
+
+> ## Fixing up remote settings {.challenge}
+>
+> It happens quite often in practice that you made a typo in the
+> remote URL. This exercice is about how to fix this kind of issues.
+> First start by adding a remote with an invalid URL:
+>
+> ~~~ {.bash}
+> git remote add broken https://github.com/this/url/is/invalid
+> ~~~
+>
+> Do you get an error when adding the remote? Can you think of a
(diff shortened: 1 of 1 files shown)


iglpdc:
Thanks @lesteve : looks good! A couple of requests, since his is a slightly more advance challenge, although I think it's totally pertinent (it happens often during the workshops that someone mistypes the URL):

Could you add a line specifying how to enter the wrong URL. E.g.:

~~~
git remote add broken https://gitbuh.com/vlad/palnets.git 
~~~

or something like that. In this case, they still have the `origin` remote working and the instructor can give a solution that works for everybody.

Also, could you also please add the "solution" under the "Remotes in Github" section in [instructors.md](instructors.md)? A link to some docs for `git remote set-url` would be enough.

Thanks again!

lesteve:
> Could you add a line specifying how to enter the wrong URL.

Sure I'll do that.

> Also, could you also please add the "solution" under the "Remotes in Github" section in instructors.md? A link to some docs for git remote set-url would be enough.

There is already a mention of git remote set-url in the [Collaborating](https://github.com/swcarpentry/git-novice/blob/gh-pages/instructors.md#collaborating) section. Do you want me to move it to the "Remotes in Github" section?

iglpdc:
> There is already a mention of git remote set-url in the Collaborating section. Do you want me to move it to the "Remotes in Github" section?

Oh, I didn't notice that :) Nah, I think it's enough with the text already there.

lesteve:
OK added the instruction to add a remote with an obviously invalid URL.

iglpdc:
Great! Thanks a lot.

PR #210, opened 2016-01-27
Learning objectives should specify a single file

MikeTheGreat:
The learning objectives make mention of 'multiple files', but the only file used is mars.txt.

COMMENT: Having a lesson that deals with multiple files, and particularly multiple files in multiple (sub)directories would be great


iglpdc:
I agree with you. I'm merging this because your version is more accurate, but I'm opening an issue to use more than one file.

PR #214, opened 2016-01-29
Instructor training

sschmeier:
I added a lesson to 06-ignore.md. I did NOT update the html.


daisieh:
This looks good to me and we can always use more challenge exercises. Thanks, @sschmeier!

PR #216, opened 2016-01-30
Improve the text in the collaboration lesson

iglpdc:
The text in the Collaborating lesson was a bit confusing on who was supposed to be to doing each of the steps. This tries to clarify a bit the commands enter by each of the roles.

I'm adding also a new challenge to repeat the whole process switching roles (you can do this simultaneously in a workshop, but I wanted to clarify that both persons should experience both roles.)

Finally, I'm cloning the Owner's repo using:

~~~
$ git clone https://github.com/vlad/planets ~/Desktop/vlad-planets
~~~

to avoid name clashes and people cloning the new repo inside their own repo.

### DIFF
--- a/08-collab.md
+++ b/08-collab.md
@@ -24,24 +24,23 @@ then select Collaborators, and enter your partner's username.
 
 ![Adding collaborators on GitHub](fig/github-add-collaborators.png)
 
-The Collaborator needs to work on this project locally. He or she should `cd` to another directory
-(so `ls` doesn't show a `planets` folder),
-and then make a copy of the Owner's repository:
+The Collaborator needs to download a copy of the Owner's repository to her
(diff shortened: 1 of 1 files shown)


daisieh:
I always suggest that the collaborator make a directory on the desktop named after the owner and then clone into that the usual way. Do you think that's more or less confusing than cloning with a specific target?

iglpdc:
For the point of view of the computer is exactly the same and I guess both are ok. I think this is a bit easier in the sense of having less room for error (in a big workshop someone is going to forget to cd to the directory after creating it) and a bit cleaner in the sense that you end up with both repos (yours and your collaborator) at the same level in your Desktop.

iglpdc:
@daisieh could you take a look at this? I think it's ready, but don't want to merge my own PR :)

daisieh:
This looks good to me, merging.

PR #220, opened 2016-01-31
Version Control History Improvements

MarkWheelhouse:
adds some additional content and time-scales to the discussion of the History of version control systems.


iglpdc:
I like it, @MarkWheelhouse. Thank you: I'm merging it.

PR #223, opened 2016-02-01
Add challenge : detached head

amueller:
This is my added challenge for the instructor training.
When I ran a short version of this class, nearly all of my students detached their had at some point, and then needed help recovering. Maybe this challenge can help.

### DIFF
--- a/05-history.md
+++ b/05-history.md
@@ -148,6 +148,19 @@ we can use a commit identifier instead:
 $ git checkout f22b25e mars.txt
 ~~~
 
+> ## Reattaching your head {.callout}
+> Above we used
+>
+> ~~~ {.bash}
+> $ git checkout f22b25e mars.txt
+> ~~~
+>
+> to revert mars.txt to its state after the commit f22b25e.
+> If you forget `mars.txt` in that command, git will tell you that "You are in
+> 'detached HEAD' state." In this state, you shouldn't make any changes.
+> You can fix this by reattaching your head using ``git checkout master``
+
+
 It's important to remember that
 we must use the commit number that identifies the state of the repository
 *before* the change we're trying to undo.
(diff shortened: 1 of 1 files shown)


iglpdc:
Thanks, @amueller. What do you think in adding this as a callout instead of a challenge? We could add it after the lines checking out the old file in the lesson. I think that having the callout right after the cause of the detached head would be more useful in the workshop.

We used to have a version of a similar callout (see `git rev-list --all | xargs git grep detached | grep 02-collab.md`), but the problem was that it mentioned branches and this was really confusing for people. 

It is also mentioned in the instructor's guide, but it maybe good to have it here for people reading the lesson by themselves.

I like your version much better, because treats the detached head as something that happens and you need to fix, which is the typical case for novice use.

If we follow this track and treat the detached head as a problem in this lesson, I think it'd better if you remove the last paragraph on using various commands in the detached head state.

amueller:
One of the reasons I added this as a challenge is because this PR is part of the instructor training, which asked for adding a challenge.
Callouts are the things with the blue pins? Having it as a note and removing the "exploration" part makes sense to me.

There is no material on branches in this lesson as far as I could tell, and the actual meaning of what this version of "checkout" does is not actually explained. So the "just fix it" approach made the most sense to me. It was the approach we took in an (admittedly brief) session I taught (after realizing that everybody ran into it).

iglpdc:
Ok, thanks! It's OK to move this as a callout even if it's for instructor training. 

And, yes, the callouts are the blue things with the pin. You can create them with:

~~~
> ## Callout title {.callout}
>
> Some text for the callout
>

Lesson continues here...
~~~

amueller:
I changed the challenge to a callout and moved it. I'm not sure what a good title would be for the callout. Maybe "Fixing detached HEAD"? I now used "Don't lose your HEAD", but actually I usually favr explicitness over humor [I'm German ;)]

iglpdc:
I like the title and appreciate the effort in using humor :). Thanks a lot! I'm merging.

PR #232, opened 2016-02-11
Add revert commit challenge

stevenvandervalk:
My Software Carpentry qualification addition.

### DIFF
--- a/05-history.md
+++ b/05-history.md
@@ -235,6 +235,18 @@ moving backward and forward in time becomes much easier.
 > 5. Both 2 & 4
 
 
+> ## Reverting a commit {.challenge}
+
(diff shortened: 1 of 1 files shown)


iglpdc:
Thanks @stevenvandervalk. We should redo this topic at some point, I think it's a bit confusing.

I like your challenge, but introducing `git revert` raises the question of which is the difference between `git checkout <commit-id> file` and `git revert <commit-id>`. Could you stress a bit more the difference between yours and the previous challenge?

stevenvandervalk:
@iglpdc agreed.  Sure, just to confirm - I should stress the difference between undoing an _actual _commit__ vs undoing just working changes?

iglpdc:
Thanks @stevenvandervalk! Merging this...

PR #218, opened 2016-01-30
Update some references across topics

iglpdc:
Also wrap all the file to 80 characters.

PR #233, opened 2016-02-12
Added an additional hosting option

burkesquires:
Adding GitLab as an addition possible public hosting option.


iglpdc:
Thanks for your PR @burkesquires, but this was included in #218.

PR #234, opened 2016-02-14
Add challenge to 06-ignore

AmaliT:
A challenge added as part of SWC training


iglpdc:
Thanks @AmaliT ! I think your exercise is very similar to the one titled "Including specific files" and we don't have time enough in the lesson to include both, so I'm closing your PR without merging.

PR #236, opened 2016-02-17
Added two references to cheatsheets

anelda:
This might be useful to refer students to at the beginning of a lesson or even to print out (at least the PDF document) and hand out to students before a workshop.  The interactive document may be confusing as it uses terminology like "index" in stead of "staging area" but it does explain so in the description when hovering over.


iglpdc:
Thanks @anelda !

PR #239, opened 2016-02-18
Added exercise to conflict section

keithfma:
The new exercise is title "A typical work session". It is intended to check
learner's understanding of how to structure a work session to deal with
potential conflicts from collaborators.


iglpdc:
I like it, @keithfma ! Thanks!

PR #242, opened 2016-02-23
history: add an exercise about how concepts can be understood

zormit:
The published version seems to be based on the HTML files which is not what I expected.
I adapted the markdown file because it was newer, but I could not really test it
(the make script did not work).

### DIFF
--- a/05-history.md
+++ b/05-history.md
@@ -202,6 +202,25 @@ on the other hand,
 moving backward and forward in time becomes much easier.
 
 
+
+> ## Checking out changed files {.challenge}
+>
+> Assume you have started to work on something, like above and git says to discard thing.
+> How can the command above "git checkout -- mars.txt" be described in terms of git?
+>
+> a) take the last snapshot and overwrite my pending changes
+> b) in the history of the repository, go back one step
+> c) clear the staging area from the changes I did to mars.txt
+> d) delete everything I have done since I last committed
+>
+> How can it be described in terms of an editor, like word?
+>
+> a) "undo" my last changes (where redo is possible)
+> b) close the file without saving
+> c) overwrite the file I am working on with an older backup
+> d) create a copy of the file I am working on and then c)
+>
+>
 > ## Recovering Older Versions of a File {.challenge}
 >
 > Jennifer has made changes to the Python script that she has been working on for weeks, and the


05-history.md
```diff
@@ -202,6 +202,25 @@ on the other hand,
 moving backward and forward in time becomes much easier.
 
 
+
+> ## Checking out changed files {.challenge}
+>
+> Assume you have started to work on something, like above and git says to discard thing.
```
davis68: Suggested rewording:  `Suppose that you have changed a working file under version control, but have not yet added or committed those changes.  If you now run the command `git checkout -- mars.txt`, what actions does Git take?`

05-history.md
```diff
@@ -202,6 +202,25 @@ on the other hand,
...
+> How can the command above "git checkout -- mars.txt" be described in terms of git?
+>
+> a) take the last snapshot and overwrite my pending changes
+> b) in the history of the repository, go back one step
+> c) clear the staging area from the changes I did to mars.txt
+> d) delete everything I have done since I last committed
+>
```
davis68: Possible rewording:

a)  takes the last commit (snapshot) and reverses any changes made in the working area since that commit.
b)  goes back one step in the history of the repository  (can test for understanding of *where HEAD is* v. *moving HEAD around*)
c)  clears the staging area of any changes to `mars.txt`    (a v. c tests for orthogonality of *things under version control* and *things which happen to share the same folder but aren't all necessarily under version control*)

Is a different from d?

05-history.md
```diff
@@ -202,6 +202,25 @@ on the other hand,
...
+> How can it be described in terms of an editor, like word?
+>
+> a) "undo" my last changes (where redo is possible)
+> b) close the file without saving
+> c) overwrite the file I am working on with an older backup
+> d) create a copy of the file I am working on and then c)
+>
```
davis68: The trouble here is which concept you are mapping commits to.  Is it *saving*, *version management* (like in Word), or simply *fundamental actions in the undo list*?  Since the mapping isn't unique, I think that this question runs the risk of confusing learners by creating faulty analogies.

05-history.md
zormit: Oh I'm sorry, that looks like some draft-y version was commited by me. The sentence I wrote there doesn't really make sense. You rewording sounds go to me, will use it in my next update of this PR.

05-history.md
zormit: The comments in parentheses are just the reasoning why you would keep an answer, aren't they?

After starting this comment, I realized it might make sense to re-work the question as a whole. I will push a new PR soon and also consider your suggestions.

05-history.md
zormit: Yeah you're right, that might be confusing. My point here was (as well in the first git-concepts-related part), that you could actually loose stuff when using checkout on changes that were never commited. So the correct answers would be b) and c). Maybe I'll find a way to make it less confusing.

iglpdc:
Hi @zormit. Thanks for your PR. I like the spirit of the exercise, but I think you need to fix the issues that @davis68 pointed out. I'm closing this PR and wait for a new one from you.

PR #244, opened 2016-02-29
Update 07-github.md

qingpeng:
move "Clone a remote repository" item from here to next lesson. "git clone" is introduced in the next lesson, not in this lesson.


daisieh:
Good idea. Merging.

Issue #141, opened 2015-05-11
Make a new figure for "Conflicts" topic


iglpdc:
After #97, all the figures have the same style, but `fig/conflict.svg`. I'd like to change it to look more like the rest.

@daisieh, could you make it? There ones you made are pretty nice.

daisieh:
Somehow I never got the notification on this! Yes, I can do this.

iglpdc:
That be great! thanks @daisieh.

daisieh:
Fixed in #245.

PR #245, opened 2016-03-01
Updated conflict.svg to match the other graphics

daisieh:
It's been many many moons since I agreed to do this in #141, but here it is at last.


iglpdc:
Thanks @daisieh!

PR #246, opened 2016-03-02
Bash command transition to Git setup commands

jhadwinOSU:
Using the bash command line lessons as a mental model, the way Git commands are setup should be addressed first in the "Setting Up Git" lesson instead of jumping right into entering Git commands without any transition/context.

### DIFF
--- a/02-setup.html
+++ b/02-setup.html
@@ -39,7 +39,14 @@
 </ul>
 </div>
 </section>
-<p>When we use Git on a new computer for the first time, we need to configure a few things. Here’s how Dracula sets up his new laptop:</p>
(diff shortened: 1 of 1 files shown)


daisieh:
This seems very reasonable to me. Merging.

iglpdc:
Hi @jhadwinOSU. Thanks for your contribution. I've reverted the merge because we only make changes in the markdown files and not in the html files, so this breaks our current workflow. Please see [more details here](https://github.com/swcarpentry/git-novice/blob/gh-pages/CONTRIBUTING.md).

Could you please remake your pull request, but introducing the changes in the `02-setup.md` file? You don't need to modify `02s-setup.html` after that.

Thanks!

PR #248, opened 2016-03-03
Reorder setup info: need/explanation/example

jhadwinOSU:
Explain the need for configurations at the beginning. Then introduce the way we would write a Git command on the command line before going into an example of the commands without any context like it was before.


daisieh:
This looks fine to me. Thanks!

PR #249, opened 2016-03-04
Edits to git configuration for editors

matthew-brett:
* Update sublime text commands on Windows;
* Add vim / emacs.

### DIFF
--- a/02-setup.md
+++ b/02-setup.md
@@ -39,12 +39,13 @@ He also has to set his favorite text editor, following this table:
 | nano               | `$ git config --global core.editor "nano -w"`    |
 | Text Wrangler      | `$ git config --global core.editor "edit -w"`    |
 | Sublime Text (Mac) | `$ git config --global core.editor "subl -n -w"` |
-| Sublime Text (Win) | `$ git config --global core.editor "'c:/program files/sublime text 2/sublime_text.exe' -w"` |
+| Sublime Text (Win, 32-bit install) | `$ git config --global core.editor "'c:/program files (x86)/sublime text 3/sublime_text.exe' -w"` |
+| Sublime Text (Win, 64-bit install) | `$ git config --global core.editor "'c:/program files/sublime text 3/sublime_text.exe' -w"` |
 | Notepad++ (Win)    | `$ git config --global core.editor "'c:/program files (x86)/Notepad++/notepad++.exe' -multiInst -notabbar -nosession -noPlugin"`|
 | Kate (Linux)       | `$ git config --global core.editor "kate"`       |
 | Gedit (Linux)      | `$ git config --global core.editor "gedit -s -w"`   |
-
-
+| emacs              | `$ git config --global core.editor "emacs"`   |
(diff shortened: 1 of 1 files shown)


iglpdc:
Thanks @matthew-brett. 

I'd rather leave out vim and emacs. Most of our learners will not use them ever and the idea of this table is not to provide an exhaustive list, but give the configuration for the editors we recommend in our workshop setup page.

gvwilson:
On the other hand, people who are already using the shell may well have 
one of these as their editor...

iglpdc:
Times when people use the real stuff are gone :) We could leave emacs and vim to cover the case of previous use, working in a remote computer without a graphical interface or restricted permissions, and historical reasons. 

I just don't want the list to be too long that it becomes too hard to maintain up to date.

matthew-brett:
I put them in because, in my classes, there are always a few people that are already using emacs or vim.

When we're teaching a semester-long class, we spend some time talking about the importance of choosing a good text editor, and we recommend [vim or emacs](http://practical-neuroimaging.github.io/choosing_editor.html) - but of course it wouldn't make sense to start learning either of these at a software carpentry workshop.

matthew-brett:
Also - I doubt the vim or emacs commands will change much or at all over time.

iglpdc:
Ok, good enough. I'm merging then.

mkcor:
@iglpdc @matthew-brett Shouldn't the accompanying HTML file ship with the PR? Run
```
$ make preview
```
to update `02-setup.html`. Right?

mkcor:
Okay: https://github.com/swcarpentry/git-novice/blob/gh-pages/CONTRIBUTING.md#getting-started
Got it!
I only found the link to the guidelines for contributing once I created my own PR (I admit I could have found it, had I looked for one extra minute).
Sorry for the noise!
How frequently are the HTML files updated, please? The delta is frustrating... (Continuous integration?)

matthew-brett:
@mkcor - I had the same problem, it took me a little while to work out what to do about the HTML file, but now I look again, I'm not sure what to suggest to make it easier to find ...  Maybe something more about the preview process and how the HTML gets into the final repo in the CONTRIBUTING document?  What do you think?

iglpdc:
@matthew-brett @mkcor yeah, this is something that maintainers are discussing right now. Continuous integration is not really nice when it happens in the middle of a workshop :) The current "random" integration is not good either... We'll come out with some rules for releases soon.

mkcor:
@iglpdc Right, I agree that actual CI is not an option. I believe a middle ground with scheduled, frequent releases would be satisfactory.
@matthew-brett I would be in favour of beefing up the CONTRIBUTING document. Some details have to wait for the lesson maintainers to come up with a decision. But maybe we can already give more reassurance about the HTML files and suggest a workflow (because running `make preview` pollutes the Git repo)... Also, I must say that reading "GitHub Flow" makes me jump to the notions of 'deployable/deploying anytime', but I admit that the reference linked to does not touch on any 'deploy*'.

PR #250, opened 2016-03-04
Rewrite callout on SSH password managers

matthew-brett:
I wonder if this section may refer to a previous version of the doc
with instructions on using SSH.  In any case, it seems a bit too
detailed for the rest of the page.  I believe it is also specific to
Linux - at least, my Mac doesn't define this variable, and it does use
password management.

### DIFF
--- a/07-github.md
+++ b/07-github.md
@@ -141,16 +141,23 @@ Branch master set up to track remote branch master from origin.
 > ## Password Managers {.callout}
 >
 > If your operating system has a password manager configured, `git push` will
-> try to use it when it needs your username and password. If you want to type
-> your username and password at the terminal instead of using
-> a password manager, type:
+> try to use it when it needs your username and password.  For example, this
+> is the default behavior for Git Bash on Windows. If you want to type your
+> username and password at the terminal instead of using a password manager,
+> type:
 >
 > ~~~ {.bash}
 > $ unset SSH_ASKPASS
 > ~~~
 >
-> You may want to add this command at the end of your `~/.bashrc` to make it the
-> default behavior.
+> in the terminal, before you run `git push`.  Despite the name, [git uses
+> `SSH_ASKPASS` for all credential
+> entry](http://git-scm.com/docs/gitcredentials#_requesting_credentials), so
+> you may want to unset `SSH_ASKPASS` whether you are using git via SSH or
+> https.
+>
+> You may also want to add `unset SSH_ASKPASS` at the end of your `~/.bashrc`
+> to make git default to using the terminal for usernames and passwords.
 
 Our local and remote repositories are now in this state:
 


gvwilson:
I believe this callout was added after some difficult experiences in one 
particular workshop - Ivan, do you remember the details?  Rather than 
deleting, perhaps it should be moved to the instructors' guide 
(`instructors.md`).

wking:
On Sat, Mar 05, 2016 at 03:45:53AM -0800, Greg Wilson wrote:
> I believe this callout was added after some difficult experiences in
> one particular workshop - Ivan, do you remember the details?

Checking the Git history in case it helps jog memories:

  $ git log -p -G SSH_ASKPASS
  …
  commit 693ae0755fd7e8d60b493275705c3bf4229825e3
  Author: Raniere Silva <raniere@ime.unicamp.br>
  Date:   Sun Aug 24 12:32:29 2014 -0300

      Add note about SSH_ASKPASS

      Close #656.

  diff --git a/02-collab.md b/02-collab.md
  index cc06c6a..417ebda 100644
  --- a/02-collab.md
  +++ b/02-collab.md
  @@ -117,6 +117,21 @@ Branch master set up to track remote branch master from origin.
   ~~~
   {:class="out"}

  +> #### Password Managers
  +>
  +> If your operating system has a password manager configured, `git push` will
  +> try to use it when it needs your username and password. If you want to type
  +> your username and password at the terminal instead of using
  +> a password manager, type
  +>
  +> ~~~
  +> $ unset SSH_ASKPASS
  +> ~~~
  +> {:class="in"}
  +>
  +> You may want to add this command at the end of your `~/.bashrc` to make it the
  +> default behavior.
  +
   Our local and remote repositories are now in this state:

   <img src="img/github-repo-after-first-push.svg" alt="GitHub Repository After First Push" />
  $ git log --graph --topo-order --oneline --decorate | grep -2 693ae0
  | * | | | | |   f336215 Merge pull request #677 from r-gaia-cs/ssh
  | |\ \ \ \ \ \  
  | | * | | | | | 693ae07 Add note about SSH_ASKPASS
  | | |/ / / / /  
  | * | | | | |   d53ff84 Merge pull request #645 from r-gaia-cs/fix-gpl-name

So initial discussion in swcarpentry/bc#677 and swcarpentry/bc#656.

matthew-brett:
From the discussion, it looks like this also applies to git-bash for Windows.

But - it still seems to me that it shouldn't be in this lesson - it's relatively advanced, and doesn't apply to what we've asked the students to do, which is to use https rather than ssh.

iglpdc:
I agree with @matthew-brett. I think there is no use for this in the lesson as currently we recommend https. But as @gvwilson says, could you move it to the instructor's guide? It's still a useful piece of information to have around.

wking:
I'm less sure this is only used for Git-over-SSH.  The credential docs
make it look like a generic fallback for all credential entry [1].

[1]: http://git-scm.com/docs/gitcredentials#_requesting_credentials

matthew-brett:
Aha - that makes a lot more sense.  Now I can reproduce the problem on a Windows machine I have to hand, when doing an https push.  It is pretty annoying.  It does seem to be set by default, and the GUI dialog that comes up doesn't let you see the characters when typing your username. The dialog says (confusingly) 'OpenSSH'. 

Does this happen on Linux as well?  I wonder whether this could be better said as:

```
## Password Manager for git bash on Windows {.callout}
>
> Git bash for Windows configures a password manager by default, and `git push` will
> try to use it when it needs your username and password. If you want to type
> your username and password at the terminal instead of using
> the password manager, type:
>
> ~~~ {.bash}
> $ unset SSH_ASKPASS
> ~~~
>
> at your git bash prompt, before using `git push`.
>
> If you want to change the default behavior each time you start git bash, ask your instructor how to
> add this command to your default configuration file
```

Or - is there lesson somewhere to point to, about editing your `.bashrc`?

matthew-brett:
Sorry - still investigating.   A recent Debian and an old Ubuntu linux machine I have to hand don't set SSH_ASKPASS by default, but another CentOS machine I have, does.    I'm leaning back towards the original phrasing, finishing up with something like my last sentence, above.

wking:
On Sun, Mar 06, 2016 at 06:28:00PM -0800, Matthew Brett wrote:
> > If you want to change the default behavior each time you start git
> > bash, ask your instructor how to add this command to your default
> > configuration file
> 
> Or - is there lesson somewhere to point to, about editing your
> `.bashrc`?

There wasn't a good place to link when the callout originally landed
[1], but that may have changed in the intervening year and a half.
Regardless of whether we have something to link on the topic, I'd
rather keep a ‘.bashrc’ mention in, instead of replacing it with “ask
your instructor”.  The former is at least something you can put into a
search engine if you're reading through the web version of the lesson
on your own (something we're trying to support [2]).

[1]: https://github.com/swcarpentry/bc/pull/677#issuecomment-53554377
[2]: http://software-carpentry.org/blog/2016/03/venn-diagram.html

matthew-brett:
OK - I rewrote the text, hoping to make it clearer that this applies to https as well as ssh.

07-github.md
```diff
@@ -140,17 +140,22 @@ Branch master set up to track remote branch master from origin.
 
 > ## Password Managers {.callout}
 >
-> If your operating system has a password manager configured, `git push` will
-> try to use it when it needs your username and password. If you want to type
-> your username and password at the terminal instead of using
-> a password manager, type:
+> Some systems will open a password manager when you run `git push`.  The
```
wking: I'm not sure about “Some systems will open…”.  It's really “Some systems will set `SSH_ASKPASS` in an attempt to be helpful, which will cause Git to open…”.  How about keeping @rgaiacs' original phrasing here and just adding a [Git for Windows][1] parenthetical?

> If your system has a password manager configured (like [Git for Windows][1] [sometimes does][2]), `git push` ….


[1]: https://git-for-windows.github.io/
[2]: https://github.com/git-for-windows/build-extra/blob/4c8d998fbb6a25e25fa16b8ba79e8efde2ffbdec/git-extra/env.sh#L4-L16

07-github.md
```diff
@@ -140,17 +140,22 @@ Branch master set up to track remote branch master from origin.
...
 >
 > ~~~ {.bash}
 > $ unset SSH_ASKPASS
 > ~~~
 >
-> You may want to add this command at the end of your `~/.bashrc` to make it the
-> default behavior.
+> in the terminal, before you run `git push`.  ([Despite the
+> name](http://git-scm.com/docs/gitcredentials#_requesting_credentials), git
+> uses `SSH_ASKPASS` whether you are using git via SSH or https).
```
wking: This addition looks good to me, although I'd suggest shifting the link:

> Despite the name, [Git prefers `SSH_ASKPASS` to terminal prompts for all credential entry][1], not just SSH, so you may want to unset it even if you are using Git over HTTPS.

[1]: http://git-scm.com/docs/gitcredentials#_requesting_credentials

matthew-brett:
I rewrote the text again, hoping to respond to your comments.

wking:
On Wed, Mar 30, 2016 at 12:10:50PM -0700, Matthew Brett wrote:
> I rewrote the text again, hoping to respond to your comments.

08536bc looks good to me.  Thanks for sticking with it :).

iglpdc:
Thanks a lot @matthew-brett ! Could you please do one little change? "git bash" -> "Git Bash".

matthew-brett:
Changed git bash to Git Bash...

iglpdc:
Thanks again!

PR #254, opened 2016-03-10
Add two challenges to 05-history.md

KristinaRiemer:
This is my exercise that is part of the requirements for SWC trainer checkout procedure.


iglpdc:
Thanks @KristinaRiemer !

PR #255, opened 2016-03-23
Add challenge to 05-history.md

stevenhwu:
This is part of the requirements for SWC trainer session.
`git log <filename>` was not mentioned in `04-changes.md`.
Since both `git diff --staged` and `git diff commitID filename` are used previously.
This should be not be a big leap to introduce `git log --switch commitID filename`

### DIFF
--- a/05-history.md
+++ b/05-history.md
@@ -289,20 +289,20 @@ Jennifer is collaborating on her python script with her colleagues and realises 
 
 > ## Checking Understanding of git diff {.challenge}
 >
-> Consider this command: `git diff HEAD~3 mars.txt`. What do you predict this command 
-> will do if you execute it? What happens when you do execute it? Why? 
+> Consider this command: `git diff HEAD~3 mars.txt`. What do you predict this command
+> will do if you execute it? What happens when you do execute it? Why?
 >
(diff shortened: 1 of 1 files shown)


iglpdc:
Hi @stevenhwu. Thanks for your PR. I like `git log <filename>` to see which commits touched particular file. I don't think the summary statistics are important enough for a novice learner. 

What do you think about keeping the first part of your challenge, where you introduce `git log <filename>`, but instead of the `--stat` you use `--patch`? (The latter allows you to see the diff for each of the commits and is the way to clarify what's is really in there for those commits with not very explanatory commit messages.)

stevenhwu:
I think this is a good idea. We probably can assume some novice users will not have very informative commit messages, and their repository is not huge. `--patch` should provide them with enough, but not overwhelming, information to explore their log.

iglpdc:
@stevenhwu do you have some time this week to finish this?  If you don't, I could add the changes to your PR.

stevenhwu:
Yes, I will work on this on Friday.

iglpdc:
Great! Thanks!

stevenhwu:
@iglpdc I update the challenge, can you please have a look and any comments/suggestions will be great. Cheers.

iglpdc:
Thanks @stevenhwu !

PR #256, opened 2016-03-23
Hotfix - fix incorrect subscript in challenges

stevenhwu:
There are three challenges where the some texts are shown in subscript and smaller font
Looks like

`Given a directory structure that looks like:` <sub>results/data results/plots</sub>

The fix placed these in a box, but can be quote as well.


iglpdc:
Thanks @stevenhwu!

PR #258, opened 2016-03-30
Update 01-basics.html

knuths:
There's a PhD comic that I like better than the one you guys use.  I updated the link when you click on the image, but I don't think I can upload an image to the fig directory in git, so the actual image on the page is the same.


daisieh:
I don't think this one works nearly as well when projected on a screen. Too much text, not enough pictures.

iglpdc:
I agree with @daisieh. Let's leave it how it was before. Thanks anywaus @knuths !

PR #264, opened 2016-04-19
SWC Instructor PR : Add challenge into 04-changes & git help info in 02-setup

yuandra:
- Added a new challenge to remove files from git repository in 04-changes
- Added an information on how people can access a list of git command & git manual in 02-setup

### DIFF
--- a/02-setup.md
+++ b/02-setup.md
@@ -76,3 +76,12 @@ same commands to choose another editor or update your email address.
 > $ git config --global --unset http.proxy
 > $ git config --global --unset https.proxy
 > ~~~
+
+> ## Git Help & Manual {.callout}
+>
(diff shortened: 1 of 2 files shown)


iglpdc:
Hi @yuandra. Thanks for your contribution. I like the commit on `git help`, but we have the file removal as part of the [discussion](https://github.com/swcarpentry/git-novice/blob/gh-pages/discussion.md#removing-a-file). The reason not to include it in the main lesson is time: if we add this exercise probably. we should explain what's the effect of removing the file with `rm <filename>` and also how to move files.

Could you drop the commit b0f819ff67bd0dcd850797de84c56c9456c3b505 for the PR? Thanks!

iglpdc:
Hey @yuandra. I've added your changes in a new PR (#298) to be able to pick up only the changes I mentioned above. The commit is still on your name. Thanks for your contribution!

PR #266, opened 2016-04-22
Added understanding HEAD as a learning objective and text to support it.

justbennet:
I grepped HEAD in previous sections, and this is it's first appearance.  To
my mind, it should be explicitly introduced, because it is a source of
some confusion.  I added the learning objective and the text to support
it to the top of the lesson.  I tried to merge without seam into what
was already there.

### DIFF
--- a/05-history.md
+++ b/05-history.md
@@ -12,11 +12,8 @@ minutes: 25
 > *   Restore old versions of files.
 
 As we saw in the previous lesson, we can refer to commits by their
-identifiers, but even the short ones do not spring trippingly from the tongue.
-Much of the time, we'll be working on some revisions, and when we want to
-refer back to something in a previous commit, it might be nice not to have to
-look up the identifier.  The clever Git people thought so, too, so you can
-refer to the _most recent commit_ by using `HEAD`.
+identifiers.  You can refer to the _most recent commit_ of the working
+directory by using the identifier `HEAD`.
 
 We've been adding one line at a time to `mars.txt`, so it's easy to track our
 progress by looking, so let's do that using our `HEAD`s.  Before we start,
--- a/05-history.md
+++ b/05-history.md
@@ -6,9 +6,59 @@ minutes: 25
 ---
 > ## Learning Objectives {.objectives}
 >
+> *   Learn what the HEAD of a repository is and how it's useful
 > *   Identify and use Git commit numbers.
 > *   Compare various versions of tracked files.
 > *   Restore old versions of files.
+
+As we saw in the previous lesson, we can refer to commits by their
+identifiers, but even the short ones do not spring trippingly from the tongue.
+Much of the time, we'll be working on some revisions, and when we want to
+refer back to something in a previous commit, it might be nice not to have to
+look up the identifier.  The clever Git people thought so, too, so you can
+refer to the _most recent commit_ by using `HEAD`.
+
+We've been adding one line at a time to `mars.txt`, so it's easy to track our
+progress by looking, so let's do that using our `HEAD`s.  Before we start,
+let's make a change to `mars.txt`.
+
+~~~{.bash}
+$ nano mars.txt
+$ cat mars.txt
+~~~
+
+~~~{.output}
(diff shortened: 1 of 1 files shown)


05-history.md
```diff
@@ -6,10 +6,60 @@ minutes: 25
...
 > *   Identify and use Git commit numbers.
 > *   Compare various versions of tracked files.
 > *   Restore old versions of files.
 
+As we saw in the previous lesson, we can refer to commits by their
+identifiers, but even the short ones do not spring trippingly from the tongue.
+Much of the time, we'll be working on some revisions, and when we want to
+refer back to something in a previous commit, it might be nice not to have to
+look up the identifier.  The clever Git people thought so, too, so you can
+refer to the _most recent commit_ by using `HEAD`.
```
QuLogic: This is not strictly true; it points to where your working directory "is". Of course, without introducing branches or detached HEAD, this statement is correct for now, so I don't know whether it's better or not to be exactly correct here.

NelleV: I would not merge this. This is, as @QuLogic says, not accurate. HEAD is a pointer to the branch you are in. If you are in something that isn't a branch, you are in "detached HEAD" mode. Introducing HEAD at this point is not necessary, and might confuse students more than anything else. I think that should be moved to after introducing branches.

carpenterbennet:
The original text includes HEAD, and I was trying to include something about this mysterious entity.  Seems like this should be rejected, then, and the lesson can continue to use HEAD without any explanation.  I'm fine with that.

iglpdc:
Thanks @justbennet for your contribution and @NelleV and @QuLogic for you comments.

I see both sides. Somehow we have mention `HEAD` because we are in fact using `HEAD~n`, but it's tricky to give an accurate definition if we cannot mention branches. Another thing is that after rereading this I feel is out of order.

This is my suggestion. Starting form the original version, I think we should move the part about `HEAD` to the very end. The learners have been seeing the full commit IDs every time they do `git log` and, during workshops, it's common that people ask what these are. So it's natural to start `git diff` with the long commit IDs. Then, as we have know, we can mention the shortened commit IDs as an alternative to save typing. Then we introduce `HEAD` and the `HEAD~n`. 

To define `HEAD` we can say that it's an alias to the commit that shows on top of your `git log` (I mean the first you see in the output of `git log`) and `HEAD~n` is n under the top one that. (This last sentence should be written better).

I think this is both intuitive (because it relates to `git log` which they will typically use to find the commit) and correct (because `git log` will show you the checkout out commit on top even on a detached head).

What do you think?

05-history.md
```diff
@@ -6,10 +6,60 @@ minutes: 25
 ---
 > ## Learning Objectives {.objectives}
 >
+> *   Learn what the HEAD of a repository is and how it's useful
 > *   Identify and use Git commit numbers.
 > *   Compare various versions of tracked files.
 > *   Restore old versions of files.
 
+As we saw in the previous lesson, we can refer to commits by their
```
iglpdc: I think this is a bit verbose, could you shorten it up a bit?

05-history.md
```diff
@@ -6,10 +6,60 @@ minutes: 25
...
+identifiers, but even the short ones do not spring trippingly from the tongue.
+Much of the time, we'll be working on some revisions, and when we want to
+refer back to something in a previous commit, it might be nice not to have to
+look up the identifier.  The clever Git people thought so, too, so you can
+refer to the _most recent commit_ by using `HEAD`.
+
```
iglpdc: I don't think we need this part. Adding a definition of HEAD to the original version will be enough.

carpenterbennet:
Ivan,

I'll try to get a look at it this evening and make re-revisions based on
your comments.  It does seem like some sort of an explanation for what HEAD
is should be provided if the lesson is going to try to tell people how to
use it.

Thanks for the suggestions for improving this.  I can try to do something
with those.

-- bennet

PeteClapham:
Thank you for your help with these
Pete

Issue #211, opened 2016-01-29
We should make changes involving more than one file


iglpdc:
When we teach the add-commit cycle, we use only one file. This is not realistic and also obscures the role played by the stage. I think we should use modify that section to use more than one file.

See also #210.

amueller:
+1

iglpdc:
Addressed in part in #268

rgaiacs:
@iglpdc Can we close this?

PR #268, opened 2016-05-13
add challenge for staging multiple files

billyc:
_((note, this PR is for my licensed-instructor checkout procedure))_

Purpose of exercise: lead student through the use-case where multiple files are added to the staging area before a commit is done.

The graphic images show multiple files going into the staging area, but there is no exercise encouraging the user to do so.


iglpdc:
Thanks @billyc. This relates to #211. I'm merging this now, but maybe we should mention this as not only as an exercise, but also as part of the lesson.

Issue #212, opened 2016-01-29
Split "Exploring history" in two topics


iglpdc:
This follows up from #205. First, the title of the topic is a bit confusing: first thing coming to my mind is the topic is about `git log`. Second, we should mention `git diff --cached`. I think it's good practive to use before `git commit` when the commit is not trivial. Third, we motivate the use of a VCS for its great "undo" features, so I think we should give the topic a more prominent place in the lesson.

My proposal is split "Exploring history" in two topics. One topic would be about `git diff` and the different formats for commit IDs and including `git diff --cached`. We can called this "Comparing different versions". The other topic would be about how to undo mistakes recovering from changes that you saved but not staged, saved and staged, saved and committed. We can called this "Undoing mistakes". This would address in the more basic parts of #179.

amueller:
+1! In particular when combined with ``checkout`` I think the ``cached`` option is important (otherwise you can't see the changes done by checking out a file).

iglpdc:
I'm closing #88, but adding it here as a reference: the rewrite of this topic has to take #88 into account.

iglpdc:
Same for #179: take it into account when rewriting this topic.

kekoziar:
It seems the novice-level suggestions in this issue have been addressed in subsequent episode PRs. Other topics suggested - such as `reset` and `rebase` - through links to other issues are intermediate and advanced topics, and so are out of scope for a novice lesson. I'm relabeling the issue as intermediate, so it can easily be found if an advanced git lesson is developed.

PR #269, opened 2016-05-14
Refactor 05-history

bpteague:
This PR is in fulfillment of the new instructor checkout requirement.

I went to improve the learning objectives in 05-history and ended up refactoring the entire lesson.  The content is the same but the emphasis and order of presentation is different.  The lesson currently starts with `git diff` and teaches readers how to refer to commits using `HEAD~x` notation, but I found I didn't have any good context to attach those new facts to, and was left asking "so what?"

My refactoring moves the concept of "going back in time" to the very beginning, motivated with the same example used later on ("Oops I over-wrote a file").  It introduces commit IDs and `git checkout` pretty much simultaneously; then short IDs, `git diff` for looking at history, and `HEAD~x`.

I welcome comments and feedback!  Unfortunately the diff is so big that reading it is a PITA; I re-rendered the HTML and pushed it to my own repo.  You can see the new version at

http://bpteague.github.io/git-novice/05-history.html

### DIFF
--- a/05-history.md
+++ b/05-history.md
@@ -53,7 +53,7 @@ We will need to manufacture our own oxygen
 ~~~
 
 `git status` now tells us that the file has been changed,
-but those changes haven't been staged:
+but those changes haven't been staged to be committed:
 
 ~~~ {.bash}
 $ git status
--- a/05-history.md
+++ b/05-history.md
@@ -1,102 +1,48 @@
 ---
 layout: page
 title: Version Control with Git
-subtitle: Exploring History
+subtitle: Exploring History and Undoing Mistakes
 minutes: 25
 ---
 > ## Learning Objectives {.objectives}
 >
-> *   Identify and use Git commit numbers.
-> *   Compare various versions of tracked files.
+> *   Identify commits by their Git commit numbers.
 > *   Restore old versions of files.
+> *   Identify commits by how many commits have happened since.
+> *   Compare different versions of tracked files.
 
-If we want to see what we changed at different steps, we can use `git diff`
-again, but with the notation `HEAD~1`, `HEAD~2`, and so on, to refer to old
-commits:
-
-~~~ {.bash}
-$ git diff HEAD~1 mars.txt
-~~~
-~~~ {.output}
-diff --git a/mars.txt b/mars.txt
-index 315bf3a..b36abfd 100644
---- a/mars.txt
-+++ b/mars.txt
-@@ -1,2 +1,3 @@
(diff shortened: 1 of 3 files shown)


gvwilson:
@daisieh @iglpdc what do you think?

iglpdc:
I prefer to split this topic into two: one for exploring the history and the other for undoing mistakes (see #212 and the comment on that).

On the new "Exploring the history" we should  mention:

- `git diff`
- `git diff --cached`
- the three ways of referring to versions: first, full commit IDs; second, short commits IDs, and third, `HEAD` and `HEAD~` (learners saw the full commit in `git log`, so it's natural start with it).

On the new "Undoing mistakes" we should mention:
- `git checkout -- <filename>` to undo changes in the working dir,
- `git reset` to undo changes in the stage,
- `git checkout <commit> <filename>` to recover an old version of the file

I usually explain this part by "incrementing the scope" of the mistake, sort of like @bpteague mentions: you first make a mistake, save it but nothing else; then you make the mistake, save and stage, but not commit, and last you go all the way to actually committing your mistake. This plays very well with any story you make up.

@bpteague do you want to redo your PR to make the first part from what you have in here. I can do the "Undoing mistakes" chunk later.

05-history.md
```diff
@@ -1,102 +1,48 @@
 ---
 layout: page
 title: Version Control with Git
-subtitle: Exploring History
+subtitle: Exploring History and Undoing Mistakes
```
rgaiacs: I'm -1 to add "Undoing Mistakes". What mistakes do you want to undo?

rgaiacs:
@bpteague do you want to redo your PR to make the first part from what you have in here?

bpteague:
Hi @iglpdc and @rgaiacs - sure, I'll be happy to re-work the PR to address only "exploring history" -- and I like the "incrementing scope" layout for @iglpdc's proposed "undoing mistakes."  I'm still left at a bit of a loss for how to motivate "exploring history" -- as a new git user, why should I care?  Thoughts?

rgaiacs:
>  I'll be happy to re-work the PR to address only "exploring history"

Thanks.

> I'm still left at a bit of a loss for how to motivate "exploring history" -- as a new git user, why should I care? Thoughts?

Because history is what enables undo features. Ask for the students to think about writing a report with **never** using the undo feature: no undo when copying and past, no undo when remove a section, no undo when remove one file, ...

kekoziar:
The changes suggested were never implemented.  Since this PR is stale and several years out of date, it will be closed.

PR #271, opened 2016-05-18
Change lesson to focus on collaboration through pull requests

jstaf:
The default SWC Git materials teach that the way to collaborate with Git/GitHub is to give collaborators direct write access to repositories. This is a little dangerous, requires a lot of trust from the repository owner, and is generally not the way that collaboration is actually done on GitHub/Bitbucket/(insert hosting provider here). 

I've replaced the collaboration lesson materials with some new ones that teach how to do things via pull request. I've also added a bit of material to index.md that gives some more solid reasons for using version control.

Anyhow, you can see the preview of my changes here: http://jstaf.github.io/git-novice/

(this pull request was submitted for instructor certification requirements, but I also just think it's a good idea to teach best practices)


iglpdc:
Hi @jstaf ! I agree with most of your thoughts. No doubt the best way of collaborating is using pull requests. However we decided to skip it in the novice lesson because of the little time we have. Pull requests require teaching branches (at least if you want to do it the proper way), which are conceptually more demanding that just adding collaborator to a repo. On the other hand, our learners typically work in small projects with a trusted group of people (say a few people from the same department working on a piece of research code or writing a paper), so the security issues are less pressing. The hope is that by working first such smaller projects with this simplified workflow, they become familiar with the tools and move later to a PR-based workflow.

We plan to teach pull requests in the coming (some time this summer) git-intermediate lesson, this would be great material to include there, but I'm closing this PR without merging here.

Thanks again!

PR #274, opened 2016-05-25
 Added review section about basic workflow in Git

chrisbpawsey:
I created a summary section for in 08-collab.md about the basic git process
git add , git commit, git push, git pull 
I think it help new git users understand better there basic process and
added a suggested best practice of adding small changes instead of a single update

### DIFF
--- a/08-collab.md
+++ b/08-collab.md
@@ -100,11 +100,11 @@ Fast-forward
 >
 > ## Review of the Basic Collaborative Workflow using Git {.callout}
 >
-> A quick review of the basic collaborative workflow using is:
-> * track a change to the repository with `git add`
-> * stage the change using `git commit -m`
+> A quick review of the git commands with a basic collaborative workflow:
+> * track a changes to the repository with `git add`
+> * stage the changes using `git commit -m`
 > * upload the changes to GitHub with `git push origin master`
-> * downloand changes to our local copy of the repository using the `git push origin master`
+> * verify that the local repo is up-to-date using `git pull origin master`
 >
 > The git commands that correspond to this workflow.
 > ~~~ {.bash}
--- a/08-collab.md
+++ b/08-collab.md
@@ -97,7 +97,28 @@ Fast-forward
  1 file changed, 1 insertion(+)
  create mode 100644 pluto.txt
 ~~~
-
+>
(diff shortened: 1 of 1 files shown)


08-collab.md
```diff
@@ -97,7 +97,28 @@ Fast-forward
...
 ~~~
-
+>
+> ## Review of the Basic Collaborative Workflow using Git {.callout}
+>
+> A quick review of the basic collaborative workflow using is:
+> * track a change to the repository with `git add`
+> * stage the change using `git commit -m`
+> * upload the changes to GitHub with `git push origin master`
+> * downloand changes to our local copy of the repository using the `git push origin master`
```
iglpdc: There is a typo here `git push origin master` should be `git pull origin master`. 

Also, do you think it would be a bit clear if you split the first three steps under an item, such as "Adding your changes to the common repo" and the fourth under another item, such as "Getting other people's changes from the common repo"? With your current wording it may seem that you should pull after pushing.

iglpdc:
Thanks @chrisbpawsey ! I've added a couple of comments.

08-collab.md
chrisbpawsey: HI Ivan,
 Thanks for taking the time to review my changes.
 Yes that is a typo should be a “git pull origin master” but looking at it with fresh eyes and your feedback.
The first line should be
A quick review of the git commands with a basic collaborative workflow:

So should the first thing that some one should do is a “git pull” to update their local copy? Is there any 
harm with doing that?  
I have the git pull at the end so that the student can see that their repo is up-to-date.
Change the last line to 
“Verify that the local repo is up-to-date” using `git pull origin master`
something like that…

I’ll make some changes and resubmit.
Cheers,
Chris B.

iglpdc:
Hi, @chrisbpawsey.

Regarding the doing `git pull` first, I think I'd be better. This is actually how one can be be sure that it's adding commits into a updated copy of the repo. If you can change it (so `git pull origin master` is the first of your four steps), it' be great.

iglpdc:
Hey @chrisbpawsey. I needed to rebase your PR, so I made a new one keeping your commits and merged it (see #299) instead of merging this one. I've reorder your workflow to pull before pushing. Thanks again for your contribution!

Issue #267, opened 2016-05-03
Privacy issues and GitHub


k8hertweck:
At a lesson debriefing this morning, an instructor noted that at least one learner was uncomfortable with the email they provided in global settings for git locally being pushed to GitHub in a later lesson (meaning their email was publicly associated with their username/legal name). After setting up a GitHub account, email addresses are not automatically published, so learners may be unaware this happens later. Given our enthusiasm to convince learners to embrace reproducibility and open science, we need to remember that some may have myriad (personal and professional) reasons for not being completely comfortable with our standard practices.

The question is how to manage the problem:
- At the very least, this should be explicitly mentioned when setting up git locally
- Should this be more explicitly mentioned in the installation instructions (where learners set up GitHub accounts?)
-  There is a [solution](https://help.github.com/articles/keeping-your-email-address-private/) provided by GitHub.

I'm interested in hearing whether other instructors have encountered similar hesitation before, or what the best/simplest resolution might be.

gvwilson:
I have had a couple of instructor trainees raise similar issues - yes, 
they should be highlighted in the notes and when we are teaching.  
GitHub's solution seems a good one.

wking:
On Tue, May 03, 2016 at 08:03:58AM -0700, Greg Wilson wrote:
> GitHub's solution seems a good one.

Yeah, using a different email like username@users.noreply.github.com
seems like the best approach to me.  I'd certainly avoid telling Git
about an email that I didn't want to make public, because it's too
easy to accidentally leak (e.g. when cherry-picking a commit from a
private repo into a public repo, the cherry-picker may not think “I
wonder if I need to adjust the author/committer name/address?”).

k8hertweck:
@jduckles and @ChristinaLK, any other thoughts given discussion this morning?

ChristinaLK:
I have no thoughts, just that we should have a ready solution for those that don't want to publish their email.  The GitHub solution seems like a good one, especially since it's "officially" published by a tool that we're asking people to use.

jttkim:
On a related issue, I quite dislike being told during a course "now go to <some URL> and create an account", so dislike having to do that during the github lesson as well.

In the past, I once managed to ask participants to create a github account if necessary at the end of day 1, so they could do this in their own time in the evening, and now it's occurred to me that this could also go into the git setup instructions, so I've just made a pull request:

https://github.com/swcarpentry/workshop-template/pull/317

rgaiacs:
[...]

PR #280, opened 2016-06-13
Addressing privacy concerns of learners

k8hertweck:
Added note to both lesson (during global configuration) and instructor notes about emails being public, as well as link to GitHub resources for configuring account to keep information private, addresses #267


daisieh:
I like it.

Issue #259, opened 2016-03-30
Add reference to "the git parable" somewhere


iglpdc:
This was suggested by @matthew-brett during one of the lesson discussion sessions. Even if the whole thing may be too advanced for the scope of this lesson, I think that at least some parts could be incorporated into it and, in any case, reading it would be helpful for most instructors.

This is [the link to the original block post](http://tom.preston-werner.com/2009/05/19/the-git-parable.html).

iglpdc:
Closed by #287

Issue #189, opened 2015-11-15
A nice command-line self-paced tutorial


gvwilson:
https://github.com/jlord/git-it

iglpdc:
We could add this a list of further resources, maybe in the instructor's guide.

iglpdc:
Closed by #287

PR #287, opened 2016-06-14
resources added to instructors guide

PBarmby:
addresses #259 and #189


daisieh:
Looks good to me. Merging.

Issue #278, opened 2016-06-10
Update collaborating section to accept push access


sstevens2:
Hey all,
Taught the git section for a workshop yesterday.  Github has some updates and now when you add someone to push to your repo, they have to accept that access.  I think we should add directions for this.  We had a little trouble because some people didn't get the little bell in the corner for notifications and we had to go to https://github.com/notifications manually to accept permissions.
Sarah

iglpdc:
Sorry, @sstevens2, I'm not sure what you mean. In the lesson we ask people to add their partner as a collaborator to their repo, so the second person gets push-rights. Are you saying that this has changed?

sstevens2:
Yes, it has changed.  Before the partner automatically got access to the repo once added.  Now they (the partner) have to accept access, by going to their notifications.

iglpdc:
I see, @sstevens2 . Do you wanna submit a PR with a fix?

sstevens2:
Created a quick and simple PR to fix it.  Let me know what you think @iglpdc 
Thanks!

iglpdc:
Fixed by #228

PR #288, opened 2016-06-14
Added info about how to accept access to repo

sstevens2:
Addresses Issue #278

I could also add a pic of an empty notification page but I'm not sure if that would really add anything.

Let me know what you think!


iglpdc:
Thanks @sstevens2 ! I think it's good enough without the picture: we can always add it later if we needed it.

PR #289, opened 2016-06-14
Add note about accepting XCode license on MacOSx

### LINKED ISSUE #193: Rstudio git integration should mention XCode command-line tools bug
This is a note that we should add information on how to deal with the executable git not being accessible on macs because the XCode license needs to be accepted.

(will provide more details later)

iglpdc:
Learners on MacOSX may need to accept the XCode license prior to the first-time use of Git.

Fixes #193


iglpdc:
I'm merging this because it's trivial.

PR #293, opened 2016-06-19
Add callout explaining how the pager works

### LINKED ISSUE #284: Add a callout box or a note regarding scrolling down through git log
If students keep adding and committing they may end up with a scrollable output from git log, which is presented as a " : " and is confusing for the novice. 

Suggested hint to add to lesson:
If output of git log is stalled on : 
this is due to the less program, which makes the output of git log scrollable.
Type q to exit this screen. Type h to get help.

iglpdc:
When the output of git log is too long, some learners may get stuck
into the pager, not knowing what to do. This adds a callout explaining
the basic behavior of the pager.

Fixes #284.

### DIFF
--- a/04-changes.md
+++ b/04-changes.md
@@ -404,6 +404,18 @@ Date:   Thu Aug 22 09:51:46 2013 -0400
     Start notes on Mars as a base
 ~~~
 
+> ## Too long of a `log`? {.callout}
+>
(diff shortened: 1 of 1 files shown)


iglpdc:
@daisieh Could check this out?

daisieh:
This looks fine to me. I'll fix the conflicts and merge.

PR #300, opened 2016-06-28
Recommend against nested git repos

TomKellyGenetics:
Changed to neutral wording, rather than "bad thing" and added more explicit recommendation against nested repos.

Added `ls -a` to the question to make it easier for participants unfamiliar with the terminal to deduce `rm -rf .git` as the answer (as it will be displayed in the terminal).

Example given below on how to demonstrate how to track within nested directories. I note that inclusion of the later content (ignore) may not be suitable here.

### DIFF
--- a/03-create.md
+++ b/03-create.md
@@ -81,7 +81,7 @@ nothing to commit (create/copy files and use "git add" to track)
 
 Git repositories can interfere with each other if they are "nested" in the directory of another. We therefore advise to create each new git repository in a separate directory. Note that we can track or ignore (as discussed later) files in directories within a git repository as shown:
 ~~~ {.bash}
-touch moon phobos deimos titan    #create moon files
+touch moon phobos deimos titan    # create moon files
 cd ..                             # return to planets directory
 ls moons                          # list contents of the moons directory
 git add moons/*                   # add all contents of planets/moons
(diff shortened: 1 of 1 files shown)


03-create.md
```diff
@@ -73,7 +73,20 @@ nothing to commit (create/copy files and use "git add" to track)
...
+
+Git repositories can interfere with each other if they are "nested" in the directory of another. We therefore advise to create each new git repository in a separate directory. Note that we can track or ignore (as discussed later) files in directories within a git repository as shown:
+~~~ {.bash}
+touch moon phobos deimos titan    # create moon files
+cd ..                             # return to planets directory
+ls moons                          # list contents of the moons directory
+git add moons/*                   # add all contents of planets/moons
+git status                        # show moons files in staging area
+git commit -m "add moon files"    # commit planets/moons to planets git repo
+"moons/*" >> .gitignore           # ignore all files within moons
```
petebachant: "moons/*" >> .gitignore doesn't work for me with Bash on Linux.

I would recommend

    echo moons >> .gitignore

rather than use a wildcard also.

rgaiacs:
@TomKellyGenetics Thanks for your work. @iglpdc We are going to migrate this pull request to the new template and open a new pull request.

PR #301, opened 2016-06-29
Configuration with Text Editors

TomKellyGenetics:
Changes to text editor configuration  …
Updated config to recommend github account as user.name

Added Scratch (Linux) as a text editor for Linux users, this is the default editor for Elementary OS.

Noted that Vim is the default editor, previous Vim users have asked helpers to configure it instead of nano (which we usually demonstrate in the workshop). Included instructions for novices to exit Vim if configuration unintentionally left on default.

### DIFF
--- a/02-setup.md
+++ b/02-setup.md
@@ -55,7 +55,12 @@ He also has to set his favorite text editor, following this table:
 
 It is possible to reconfigure the text editor for Git whenever you want to change it.
 
-Note that Vim is the default editor for Git so this is only needed if you have changed it already. If you haven't used Vim before and wish to exit a session, type <kbd>Esc</kbd> then <kbd>:q!</kbd> and <kbd>Enter</kbd>.
+> ## Exiting Vim {.callout}
+>
+> Note that `vim` is the default editor for for many programs, if you haven't used `vim` before and wish to exit a session, type <kbd>Esc</kbd> then <kbd>:q!</kbd> and <kbd>Enter</kbd>.
+>
+
+
 
 The four commands we just ran above only need to be run once: the flag `--global` tells Git
 to use the settings for every project, in your user account, on this computer.
--- a/02-setup.md
+++ b/02-setup.md
@@ -46,7 +46,7 @@ He also has to set his favorite text editor, following this table:
 | Sublime Text (Mac) | `$ git config --global core.editor "subl -n -w"` |
 | Sublime Text (Win, 32-bit install) | `$ git config --global core.editor "'c:/program files (x86)/sublime text 3/sublime_text.exe' -w"` |
 | Sublime Text (Win, 64-bit install) | `$ git config --global core.editor "'c:/program files/sublime text 3/sublime_text.exe' -w"` |
-| Notepad++ (Win)    | `$ git config --global core.editor "'c:/program files (x86)/Notepad++/notepad++.exe' -multiInst -notabbar -nosession -noPlugin"`|
+| Notepad++ (Win)    | `$ git config --global core.editor "'c:/program files (x86)/Notepad++/notepad++.exe' -multiInst -notabbar -nosession -noPlugin"` `$ alias notepad="'c:/program files (x86)/Notepad++/notepad++.exe'"`|
 | Kate (Linux)       | `$ git config --global core.editor "kate"`       |
 | Gedit (Linux)      | `$ git config --global core.editor "gedit -s -w"`   |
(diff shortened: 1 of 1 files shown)


02-setup.md
```diff
@@ -32,7 +32,8 @@ $ git config --global color.ui "auto"
 
 (Please use your own name and email address instead of Dracula's. 
 This user name and email will be associated with your subsequent Git activity, 
-which means that any changes pushed to GitHub in a later lesson will include this information. 
+which means that any changes pushed to GitHub in a later lesson will include this information.
+We recommend setting up an account on [GitHub](https://github.com) if you haven't already and using this as your user.name so your computer works with this website in the future (including later in this lesson).
```
iglpdc: I'm not sure I understand this. git's user.name is the "real name" of the person and not GitHub's login name. While it's possible to use GitHub's username as the name is your commits, in my experience this creates a lot of problems with attribution, and some people, including SWC, prefer having contributors for a real name in their commits

02-setup.md
```diff
@@ -48,9 +49,17 @@ He also has to set his favorite text editor, following this table:
 | Notepad++ (Win)    | `$ git config --global core.editor "'c:/program files (x86)/Notepad++/notepad++.exe' -multiInst -notabbar -nosession -noPlugin"`|
 | Kate (Linux)       | `$ git config --global core.editor "kate"`       |
 | Gedit (Linux)      | `$ git config --global core.editor "gedit -s -w"`   |
+| Scratch (Linux)       | `$ git config --global core.editor "scratch-text-editor"`  |
```
iglpdc: I'm not familiar with Scratch. I'm happy to include it if it's somehow popular among learners. We don't want to turn this section in a exhaustive documentation list of all editors.

02-setup.md
```diff
@@ -48,9 +49,17 @@ He also has to set his favorite text editor, following this table:
...
 | emacs              | `$ git config --global core.editor "emacs"`   |
+
+It is also possible to configure "Vim" as the text editor for Git if you want to change it back. Note that Vim is the default editor for Git so this is only needed if you have changed it already.
```
iglpdc: I prefer to leave the table as it was (not making `vim` a new row). Instead of mentioning that `vim` is the default, we could mention that you can change the configuration (the editor and all the rest) as many times as you want. This is a question that shows up in the workshops and some people is not used to be able to configure computer programs as they please (they tend to worry about picking the "wrong" editor and having to stuck with it for life).

but I agree that it's very, very important to include the

02-setup.md
```diff
@@ -48,9 +49,17 @@ He also has to set his favorite text editor, following this table:
...
+
+| Default Editor             | Configuration command                            |
+|:-------------------|:-------------------------------------------------|
 | vim                | `$ git config --global core.editor "vim"`   |
 
+If you haven't used Vim before and wish to exit a session, type <kbd>Esc</kbd> then <kbd>:q!</kbd> and <kbd>Enter</kbd>.
```
iglpdc: This is a very, very welcomed addition! Could you make it a callout? `vim` is not only the default for `git`, but for some other programs, so this is useful to have stressed.

Also, we prefer to avoid the html in markdown, so it'd be great if you used just back ticks.

iglpdc:
Thanks @TomKellyGenetics ! Could you check out my comments?

02-setup.md
TomKellyGenetics: Thanks, this was a misunderstanding on my part on how Git needs to be configured.

02-setup.md
TomKellyGenetics: Use of scratch is likely rare at this point, although it is becoming more popular as the default in elementary OS (a linux distro I recommend to beginners). I have mainly included it because the terminal command for Scratch is not immediately obvious to users of the GUI interface.

I realise this could snowball into a long list of everyone's favourites but this is one I've had trouble configuring myself and one I would recommend to new users (rather than `vim` or `emacs`).

02-setup.md
TomKellyGenetics: Noted, the preference for backticks has also been discussed here: https://github.com/swcarpentry/shell-novice/pull/421

TomKellyGenetics:
Thanks for the feedback @iglpdc, I'll change accordingly.

rgaiacs:
@iglpdc Could you merge https://github.com/swcarpentry/git-novice/pull/301 and close this one?

iglpdc:
Merged by #320

PR #303, opened 2016-06-30
Added Branches lesson

TomKellyGenetics:
I've written a branches lesson which would ideally be used after Conflicts (or possibly after Ignore). As it's a more advanced topic. I've put it as a supplementary lesson for us to discuss and to use to demonstrate extension features of Git.

I note that this will be a controversial topic to present to absolute beginners but it would benefit many users to be exposed to this as they will be joining an existing company or institutional repoositories using branches.


iglpdc:
We have no time to talk about branches in this lesson, but this material could be reused for the future. Intermediate lesson.

TomKellyGenetics:
I originally wrote this as notes for myself to demonstrate branches at the end. We realise that time is pressing and it would not be suitable for absolute beginners. Some learners at our workshop were members of our research group who we wanted to join projects using branches so this was mainly for their benefit. In the interest of time we're likely to cover this as a separate "Mozilla Study Group" session instead.

I'm happy for this to be used for an intermediate lesson.

kekoziar:
PRs labeled _intermediate_ and _out of scope_ can still be accessed if closed.

PR #312, opened 2016-07-20
Updates to lesson 4

dopplershift:
This fixes a few things that jumped out at me while reading through lesson 4 on tracking changes. 

I'm doing this as part of my checkout procedure for becoming an instructor.


iglpdc:
Nice @dopplershift ! Thank you.

PR #285, opened 2016-06-13
Addresses Issue 283 to double check the wd 

gallington:
Added a few lines to check the working directory is reset to the correct level after the last section.

Issue #283, opened 2016-06-13
Add clarification of working directory to start of Tracking Changes section


gallington:
In Lesson: Version Control with git
Creating Repository ---> Tracking Changes
--at end of Creating Repository section you create a new dir called moons. It is unclear whether this is to be the new wd for the next step (Tracking Changes) or if it should be moved back to planets. 
Add direction as to which working directory you should be in: planets or moons.
Should moons subdirectory still exist? Or should this be deleted altogether?

iglpdc:
Oh, @gallington, I see your point. It's true that maybe confusing. It'd be very simple to fix adding some text at the beginning the next topic, in the sentence: 

"Let's create a file called `mars.txt` that contains some notes ... ".

Maybe:

 "Let's go back to the planets directory (`$cd ~/Desktop/planets$`) and create a file called `mars.txt` that contains some notes ..."

@gallington, do you wanna send a fix? :)

brainwane:
I think it's a great idea to have a step here where we make sure that the learner is in the correct directory -- do we need to take anything else *out* of this episode to make room for this step? I ask this particularly because the earlier episodes do not use `pwd`, so we're introducing it for the first time here, which may end up taking 30 seconds to a minute of explanation.

PR #330, opened 2016-08-27
Add clarification of working directory to start of Tracking Changes section

rgaiacs:
Cover #283 and #285.

### DIFF
--- a/_episodes/04-changes.md
+++ b/_episodes/04-changes.md
@@ -17,6 +17,21 @@ keypoints:
 - "Always write a log message when committing changes."
 ---
 
+First let's make sure we're still in the right directory.
+You should be in the `planets` directory.
+
+~~~
+$ pwd
+~~~
+{: .bash}
+
+If you are still in `moons` navigate back up to `planets`
+
+~~~
+$ cd ..
+~~~
+{: .bash}
+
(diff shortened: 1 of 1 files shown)


brainwane:
I'm wondering whether we should add sample expected output:

`~~~
/home/user/planets
~~~
{: .output}`

But since every learner will have a different path, that might cause more confusion than it addresses.

rgaiacs:
> I'm wondering whether we should add sample expected output, e.g. /home/user/planets with the {: .output} formatting. But, since every learner will have a different path, that might cause more confusion than it addresses.

@brainwane I didn't understand how your comment is related with my pull request.

brainwane:
Hi @rgaiacs -- sorry for not being clear. I was suggesting that, since your pull request includes a suggested addition where the instructor demonstrates an input (invoking `pwd` on the command line), it might be a good idea to also include an addition of sample output. But I'm also not absolutely certain of this and willing to be told "bad idea" or "let's put that in a separate issue". :)

daisieh:
I think this looks good as-is. Usually the instructor is live-coding this as it happens, so the class would see the output.

brainwane:
OK, thanks @daisieh! Suggestion withdrawn. :)

PR #337, opened 2016-08-27
Changed lesson 06-ignore.md to improve exercise

aheilmaier:
Changed exercise because of little confusion.

### DIFF
--- a/_episodes/06-ignore.md
+++ b/_episodes/06-ignore.md
@@ -155,21 +155,21 @@ nothing to commit, working directory clean
 > How would you ignore only `results/plots` and not `results/data`?
 >
 > > ## Solution
-> > 
-> > As with most programming issues, there are a few ways that you
-> > could solve this. If you only want to ignore the contents of 
-> > `results/plots`, you can change your `.gitignore` to ignore 
-> > only the `/plots/` subfolder by adding the following line to 
-> > your .gitignore: 
-> > 
-> > `results/plots/` 
-> > 
-> > If, instead, you want to ignore everything in `/results/`, but wanted to track
-> > `results/data`, then you can add `results/` to your .gitignore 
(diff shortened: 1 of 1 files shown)


rgaiacs:
I believe that [**-pattern format](https://git-scm.com/docs/gitignore#_pattern_format) doesn't belong to the novice lesson because it is complex:

> Two consecutive asterisks ("**") in patterns matched against full pathname may have special meaning:
>
> - A leading "**" followed by a slash means match in all directories. For example, "**/foo" matches file or directory "foo" anywhere, the same as pattern "foo". "**/foo/bar" matches file or directory "bar" anywhere that is directly under directory "foo".
> - A trailing "/**" matches everything inside. For example, "abc/**" matches all files inside directory "abc", relative to the location of the .gitignore file, with infinite depth.
> - A slash followed by two consecutive asterisks then a slash matches zero or more directories. For example, "a/**/b" matches "a/b", "a/x/b", "a/x/y/b" and so on.
>
> Other consecutive asterisks are considered invalid.

I suggested @aheilmaier to replace it with a shell glob.

> Otherwise, Git treats the pattern as a shell glob suitable for consumption by fnmatch(3) with the FNM_PATHNAME flag: wildcards in the pattern will not match a / in the pathname. For example, "Documentation/*.html" matches "Documentation/git.html" but not "Documentation/ppc/ppc.html" or "tools/perf/Documentation/perf.html".

Paths underneath are cover during the lesson by pattern that ends with a slash.

> If the pattern ends with a slash, it is removed for the purpose of the following description, but it would only find a match with a directory. In other words, foo/ will match a directory foo and paths underneath it, but will not match a regular file or a symbolic link foo (this is consistent with the way how pathspec works in general in Git).

iglpdc:
Thanks @aheilmaier! Yes, I agree `**`was confussing: it's better now.

PR #349, opened 2016-10-11
Update 13-hosting.md

pshellito:
Tried to fix broken link to page source... still not working. I shouldn't have committed yet, I'll take another look (I'm new to this)...

### DIFF
--- a/_episodes/13-hosting.md
+++ b/_episodes/13-hosting.md
@@ -42,7 +42,7 @@ communities already using the same service.
 
 As an example, Software Carpentry [is on
 GitHub](https://github.com/swcarpentry/) where you can find the [source for this
-page](https://github.com/swcarpentry/git-novice/blob/gh-pages/13-hosting.md).
+page](https://github.com/swcarpentry/git-novice/blob/gh-pages/_episodes/13-hosting.md).
 Anyone with a GitHub account can suggest changes to this text.
 
 Using large, well-established services can also help you quickly take advantage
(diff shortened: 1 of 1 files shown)


_episodes/13-hosting.md
```diff
@@ -42,7 +42,7 @@ communities already using the same service.
 
 As an example, Software Carpentry [is on
 GitHub](https://github.com/swcarpentry/) where you can find the [source for this
-page](https://github.com/swcarpentry/git-novice/blob/gh-pages/13-hosting.md).
+page](https://github.com/swcarpentry/git-novice/blob/gh-pages/_episodes/13-hosting.md).
```
iglpdc: Good catch! But we prefer to use the variables defined in `_config.yml` for this type of URLs, so we don't get broken links so when we move things around :)

`{{ page.root }}` equals the root dir of the repo. So you should do `...[source for this page]({{page.root}}/_episodes/13-hosting.md`

_episodes/13-hosting.md
```diff
@@ -42,7 +42,7 @@ communities already using the same service.
 
 As an example, Software Carpentry [is on
 GitHub](https://github.com/swcarpentry/) where you can find the [source for this
```
iglpdc: You could also change this one, if you don't mind: `https://github.com/swcarpentry`-> ` {{ swc_github }}` (see again `_config.yml`.

iglpdc:
Thanks @pshellito !

pshellito:
@iglpdc looks like it's still not working, unfortunately. Maybe it just need some time to re-run jekyll.

PR #351, opened 2016-10-14
Fix bugs in create.md formatting

daisieh:
There were some formatting glitches in 03-create.md. I also added a bit more explanation for why it’s bad to have nested repos.


maglet:
Thanks for fixing this! I'm teaching this on Sunday, would be nice if it gets merged before then

iglpdc:
Thanks!

PR #366, opened 2016-11-18
update gedit config

### LINKED ISSUE #174: `gedit -s` not working properly
I set up gedit to be the default editor with the -s option, but it is not working

This is what I get when I do `gedit --version`:
`gedit - Version 2.28.4`

pierre-haessig:
It seems that gedit command line interface is a moving target. After the fix #174 from January 2016, now it's again broken! On Ubuntu 16.04 with gedit (v 3.18) and Debian testing (v 3.22), the `gedit -w -s` call yields:

    Failed to register: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod:
    Method "DescribeAll" with signature "" on interface "org.gtk.Actions" doesn't exist

I found a report for this issue (https://bugzilla.gnome.org/show_bug.cgi?id=742469) from Jan 2015. Apparently for gedit 3.14. It's not yet fixed, but a workaround is suggested, to replace the '-s' (standalone) option by 'new-window'. New git config looks like:

    $ git config --global core.editor "gedit --wait --new-window"

I've tested one commit message (with one extra gedit window previously opened to check the new-window option) and it seems to work fine. I think 02-setup.md should be updated accordingly.


iglpdc:
Ok @pierre-haessig , thanks a lot!

PR #371, opened 2016-12-14
Tracking Changes: 2 notes about directories in Git

maxim-belkin:
Added two notes about directories in Git.
1. Directories on their own are not tracked.
2. When you do `git add <directory>`, you effectively add all the files
within that directory.


maxim-belkin:
Forgot to mention that this PR should be considered as a part of my Instructor Training Checkout Procedure

maxim-belkin:
One more idea: the note could compare the addition of an empty file and of an empty directory.

iglpdc:
Thanks @maxim-belkin!

PR #373, opened 2016-12-22
Changed the description of avoidance strategies 

AmericanUmlaut:
I found the previous description a bit unclear.

"Version control’s ability to merge conflicting changes is another reason users tend to divide their programs and papers into multiple files instead of storing everything in one large file."

Rather than the ability to merge conflicts, it is the undesirability of having conflicts in the first place that leads users to prefer more, smaller files over fewer, larger ones.


iglpdc:
Thanks @AmericanUmlaut. I think it's clearer now.

PR #374, opened 2016-12-23
Create a solution for exercise 09-conflict

weiapplele:
The solution for "how to resolve a conflict" exercise for lesson 09-conflict


iglpdc:
I think for this exercise the solution would be just pointing to the learner to repeat the work done in this episode.

ErinBecker:
I agree with @iglpdc's suggestion to avoid introducing `git stash` here as unnecessary complexity. This is beyond the current level of the lesson, although it would be a great addition to an intermediate Git lesson.

PR #377, opened 2016-12-31
Avoid early mention of `mars` repository

tbarron:
At this point in the lesson, the `mars` repository has not been mentioned. Rather, it's the `moons` repo that was created nested in the example.


maxim-belkin:
+1, although #372 has been merged already

PR #384, opened 2017-02-01
fixing mistake in line 90

arockenberger:
'mars' must be 'moons'


rgaiacs:
Great catch!

iglpdc:
Thanks you @arockenberger!

PR #386, opened 2017-02-04
Updating mailmap and AUTHORS before release

twitwi:


twitwi:
(some context about this PR)

The authors file is used to list all authors of the lesson, which currently means "people having commited to the repo". This file is used for the release of the lessons (where we get a DOI from Zenodo, publish on our site [1], etc).

We generate the authors file from `git log` by using the commiter names. However, for a same person we have commits showing different emails, different writings of the name, a pseudo instead of the name, etc. Git has the `.mailmap` mechanism [2] exactly to solve this issue by describing how commits should be patched. This file used to be filled manually (leading to useless entries, missing ones, errors) and it is now assisted by a script, leading to some notable cleanup.

[1] https://software-carpentry.org/lessons/previous/
[2] https://git-scm.com/docs/git-shortlog#_mapping_authors

twitwi:
Ordered AUTHORS as per comment/suggestion of some maintainer.

twitwi:
@iglpdc which version would you rather have me use for the release? before or after the recent merges?

twitwi:
cc @daisieh

twitwi:
Closing in favor of #387

PR #387, opened 2017-02-08
Updating AUTHORS and mailmap

twitwi:
This is based on the latest version, with the new authors.

I'm closing #386 as it should not be merged.
The question about which version to publish remains and I'll keep both branches in the meantime.


twitwi:
With respect to #386, I made a release based on the most recent version. If it is a problem, let me know and I'll try to fix it.

In any case, the current PR is meaningful for the longer term.

daisieh:
Sorry about my complete absence from this whole thing: this is fine with me.

twitwi:
I'm merging it then, while it is fast forward.

PR #388, opened 2017-02-14
Add solutions for exercises 2 and 3 in 09-conflicts

tgamblin:
This provides solutions to two exercises on conflicts:
- resolving conflicts in image files, and
- a Parson's problem on avoiding conflicts

This does not provide a solution to the first conflict exercise, as it
would be very similar to the lesson itself.

### DIFF
--- a/_episodes/09-conflict.md
+++ b/_episodes/09-conflict.md
@@ -294,9 +294,9 @@ We removed the conflict on this line
 
 We don't need to merge again because Git knows someone has already done that.
 
-Git's ability to resolve conflicts is very useful, but conflict resolution 
-costs time and effort, and can introduce errors if conflicts are not resolved 
-correctly. If you find yourself resolving a lot of conflicts in a project, 
+Git's ability to resolve conflicts is very useful, but conflict resolution
+costs time and effort, and can introduce errors if conflicts are not resolved
+correctly. If you find yourself resolving a lot of conflicts in a project,
 consider one of these approaches to reducing them:
 
 - Try breaking large files apart into smaller files so that it is less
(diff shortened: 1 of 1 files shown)


_episodes/09-conflict.md
```diff
@@ -320,6 +320,146 @@ consider one of these approaches to reducing them:
...
+> > ~~~
+> > {: .output}
+> >
+> > We've learned that we must pull first and resolve any conflicts:
+> >
+> > ~~~
+> > $ git pull origin master
+> > ~~~
+> > {: .bash}
+> >
```
katyhuff: Maybe some transition here?

tgamblin: Updated!

katyhuff:
@iglpdc @daisieh What do you think of this? It seems like a great addition to me.

daisieh:
I'm loath to introduce the idea of committing or dealing with conflicts to binary files. I think it would be confusing to new users who are just being introduced to the concept of diffing on text files.

tgamblin:
This is the solution to an exercise that was already in the episode, so it's already introduced.  I suppose you could remove the particular exercise, but I suspect that it's something people will ask.  Is there a risk of confusing people if it's buried under an exercise?

iglpdc:
Thanks @tgamblin. I agree with @daisieh that talking about conflicts for binary files maybe a bit too much for a typical learner. However, your text is really well-written and very easy to follow. Instructors are supposed to choose which exercises are done during the workshop, so I think we should merge your PR and let to the instructor's jugedment whether is appropiate for their audience or not.

I also think that this may cover the case when people put a previous project under git, committing in the process a binary file (say some data on a spreadsheet). The spreadsheet changes, they get a conflict which they cannot fix by editing the file, they cannot pull/push anymore and they quit using Git.

I have one minor request though. Could you think in an easy way of creating the jgp (or whatever binary) files? I'm thinking that in a workshop people would not have handy jpgs to share, or they will be too big, or whatever.  It would be ideal if there was something using only command line stuff and available in all OSes.

Or maybe just taking a screenshot of the learner's computer would be enough :-)

Any ideas @katyhuff ?

iglpdc:
This could be a solution for generating a small, random file with no personal stuff on it:

```
dd if=/dev/urandom of=./mars.jpg bs=1 count=1024
```

but uses `dd`, which has the "weird" format (from Solaris?) to pass options with `=` signs, instead of the more common `--` that we teach.

tgamblin:
Would it be sufficient to check in the jpg's as part of this exercise, and just use `curl -O` to get them?

tgamblin:
Alternately, if you really want random data, you can do this without having to use `dd`

```
cat /dev/urandom | head --bytes 1024 > mars.jpg
```

iglpdc:
> head --bytes 1024 /dev/urandom > mars.jpg

I think that's perfect! All these are commands that people is familiar with from the shell lesson. Could you just add a short sentence with that when you first mention the `mars.jpg` file? Something like: "If you don't have a real picture of Mars, you can create a small binary file with these commands ..."

tgamblin:
@iglpdc: done!

iglpdc:
Very nice! Thanks a lot @tgamblin!

tgamblin:
Thanks everyone!

PR #389, opened 2017-03-04
Update 04-changes.md

abracarambar:
Add '$' character for code snippets that needs to be executed; and fix list numbering in Directories section


iglpdc:
Thanks @abracarambar!

PR #390, opened 2017-03-05
Added git show to to 05-history

leobrowning92:
Added the `git show` command to the section on reviewing history.

I tried to remove some content so as not to add more length to the lesson. The content removed was mostly repeated examples of `git diff`, which I think can be replaced without losing any knowledge of how `git diff` works and provides teachers and learners with other options for reviewing their history.


leobrowning92:
This is in response to issue #399

iglpdc:
Thanks @leobrowning92 !

PR #392, opened 2017-03-07
Gitignore notes and word-based diffing

junghans:
- Added some notes about `git://` and ssh proxy
- Added a note about `**` pattern in gitignore section
- Mention `--color-words` option in diff section, which is extremely useful for LaTeX files.

### DIFF
--- a/_episodes/04-changes.md
+++ b/_episodes/04-changes.md
@@ -474,6 +474,14 @@ Date:   Thu Aug 22 09:51:46 2013 -0400
 ~~~
 {: .output}
 
+> ## Word-based diffing
+>
+> Sometimes, e.g. in the case of the text documments a line-wise
+> diff is too coarse. That is where the `--color-words` option of
+> `git diff` comes in very useful as it highlights the changed 
+> words using colors.
+{: .callout}
+
 > ## Paging the Log
 >
 > When the output of `git log` is too long to fit in your screen,
--- a/_episodes/06-ignore.md
+++ b/_episodes/06-ignore.md
@@ -168,6 +168,9 @@ nothing to commit, working directory clean
 > > and create an exception for the `results/data/` folder.
 > > The next challenge will cover this type of solution.
 > >
+> > Sometimes the `**` pattern comes in handy, too, which matches
+> > multiple directory levels. E.g. `**/results/plots/*` would make git ignore
+> > the `results/plots` directory in any root directory.
 > {: .solution}
 {: .challenge}
 


_episodes/04-changes.md
```diff
@@ -474,6 +474,13 @@ Date:   Thu Aug 22 09:51:46 2013 -0400
 ~~~
 {: .output}
 
+> ## Word-based diffing
+>
+> Sometimes, e.g. in the case of the LaTeX documment a line-wise
```
iglpdc: documment -> document

iglpdc: There is a missing comma after "document":

"Sometimes, e.g. in the case of a LaTeX document, ..."

but I think it may even better not to mention the LaTeX example: most of our learners do not use it and this may make them think that the color diff is not useful for them.

_episodes/04-changes.md
```diff
@@ -474,6 +474,13 @@ Date:   Thu Aug 22 09:51:46 2013 -0400
...
+> diff is too coarse. That is where the `--color-words` option of
```
iglpdc: I would rephrase it to explain a bit what color-words does. From the docs: "Highlights the changed words using colors". What do you think of changing your sentence "That is where ... " to include something similar?

_episodes/02-setup.md
```diff
@@ -98,6 +98,8 @@ same commands to choose another editor or update your email address.
 > $ git config --global --unset https.proxy
 > ~~~
 > {: .bash}
+> In addition git supports the SSH protocol, which needs no proxy configuration, and
```
iglpdc: I'd prefer not to include this change. I think it's useful information for a more advanced crowd, but it may be confussing for a novice. The only reason the proxy callout is here is because it may be needed in some workshops with a weird network configuration.

_episodes/07-github.md
```diff
@@ -144,6 +144,7 @@ Branch master set up to track remote branch master from origin.
 > $ git config --global --unset https.proxy
 > ~~~
 > {: .bash}
+> Please remember that for using SSH as connection, proxy setup is not needed.
```
iglpdc: Something as above. I think mentioning details of SSH won't be useful and may be confussing for most of our learners. We have a few links to set up properly SSH in this episode and I think this is enough.

junghans:
@iglpdc: Done!

iglpdc:
Thanks a lot @junghans !

PR #395, opened 2017-03-15
Regular windows and echo for editing suggested

LiuGangKingston:
Suggestions are put into the new file suggestions.txt

### DIFF
--- a/suggestions.txt
+++ b/suggestions.txt
@@ -1,0 +1,59 @@
+
+The followings are suggested
+
+
+1) The font size of git bash should be set big enough by
+   right clicking git bash window -> Options -> Text -> Select... -> Size -> 20
+
+
+2) For Microsoft Windows system, C, D, E disks are accessible by commanding
+   cd /c/
+   cd /d/
+   cd /e/
+   respectively, in the git bash.
+
+
+3) To open one or two normal windows to list the files in the working directories normally,
+   so that participants can watch some magic shows.
+   When versions are changed back and forth, some files may appear or disappear.
+   Similar situation when branches are switched.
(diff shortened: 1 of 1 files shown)


suggestions.txt
```diff
@@ -0,0 +1,59 @@
...
+        the command line. Luckily only short messages are needed for the course. However at
+        a later time, participants should either setup favourite editor or learn to use vi
+        and/or vim.
+
+   4.2) Use echo command and stream redirection to append one line to an existing file, like
+        echo "There an apple and Newton in Mars" >> mars.txt
+        echo '!@#$%^&*()_+ "this is important."' >> mars.txt
+
+   4.3) Big text files can be editted by using method in 3).
+
```
iglpdc: We use nano to avoid `vim` complexities. We also cover `echo` and redirection, but it the shell lesson, which is a prerequisite for this one.

suggestions.txt
```diff
@@ -0,0 +1,59 @@
...
+5) Talk about using stream rediction to collect results from git log, like
+   git log > tmp01.txt
+   then use a favourite editor to explore the file for interested information.
+
+
+6) Branching should be addressed
+
+   6.0) git clone a.new.git.repository
+        then go to the directory
+   6.1) git checkout -b a.new.branch
```
maxim-belkin: I'd advocate against `checkout -b` for the learners and use the original `git branch <blah> && git checkout <blah>`. Once that is explained, it is safe to mention that you can do both of these steps at once with `checkout -b`

kekoziar:
Most of these suggestions are out of scope of the lesson.  Closing.

PR #396, opened 2017-03-22
Update 04-changes.md

mckinniburgh:
In the beginning instructions on creating and editing a file in the newly initiated repo, I have added a reminder that the bash command to open or edit text editors depends on the editor you choose. I link back to the more thorough discussion of text editors in The Unix Shell lesson (third episode). 

I find that students, particularly those new to command line interfaces, often get tripped up on opening, editing, and closing text editors. My edit is an attempt to remind students to double-check and clarify that they'll need to select their editor and the right command that goes with it.

### DIFF
--- a/_episodes/04-changes.md
+++ b/_episodes/04-changes.md
@@ -19,9 +19,9 @@ keypoints:
 
 Let's create a file called `mars.txt` that contains some notes
 about the Red Planet's suitability as a base.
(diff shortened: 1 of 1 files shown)


iglpdc:
Makes sense, the links to the shell lesson are specially helpful for people learning on their own. Thanks, @mckinniburgh!

PR #397, opened 2017-03-27
Alternative Solutions for 05-history.md

AlexSUnderwood:
I know there is currently a pull request open right now containing solutions to some of the challenges in 05-history.md, but I felt it might help to write some solutions that better fit with the "role-play" of the challenges and help explain a little bit more about what each choice in the challenge does.

This is for instructor training checkout.

### DIFF
--- a/_episodes/05-history.md
+++ b/_episodes/05-history.md
@@ -348,6 +348,30 @@ moving backward and forward in time becomes much easier.
 > 4. `$ git checkout <unique ID of last commit> data_cruncher.py`
 >
 > 5. Both 2 and 4
+>
+> > ## Solution
+> >
+> > Lets look at each command to see what will happen.
+> > 
+> > 1. `$ git checkout HEAD`
+> >
+> > When Jennifer runs this command, git will not change any of her files because this command does not specify which files she wants to checkout.
+> >
+> > 2. `$ git checkout HEAD data_cruncher.py`
+> >
+> > With this command, git will replace Jennifer's "broken" copy of data_cruncher.py with the previous commit's unmodified version.  She'll be given a "pre-break" file and have a chance to try editing it again.
+> >
+> > 3. `$ git checkout HEAD~1 data_cruncher.py`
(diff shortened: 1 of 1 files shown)


_episodes/05-history.md
```diff
@@ -348,6 +348,30 @@ moving backward and forward in time becomes much easier.
...
 >
 > 5. Both 2 and 4
+>
+> > ## Solution
+> >
+> > Lets look at each command to see what will happen.
+> > 
+> > 1. `$ git checkout HEAD`
+> >
+> > When Jennifer runs this command, git will not change any of her files because this command does not specify which files she wants to checkout.
```
maxim-belkin: When no file names are specified, `git checkout <ref>` moves the HEAD pointer to the specified <ref>. Therefore, `git checkout HEAD` does not change any files not because there are no files specified but because `git checkout` behaves differently: it moves `HEAD` to `HEAD`. `git checkout HEAD~1`, for example, will change contents of all files. I'd suggest either expanding this explanation or adding a small section mentioning that `checkout` (and `reset`) behave differently when file names are specified on the command line.

_episodes/05-history.md
```diff
@@ -348,6 +348,30 @@ moving backward and forward in time becomes much easier.
...
+> >
+> > 2. `$ git checkout HEAD data_cruncher.py`
+> >
+> > With this command, git will replace Jennifer's "broken" copy of data_cruncher.py with the previous commit's unmodified version.  She'll be given a "pre-break" file and have a chance to try editing it again.
```
maxim-belkin: "broken" is jargon. You can either use "current" or omit this word at all.
"pre-break" is jargon as well. In fact, I don't think we need the second sentence altogether.

kekoziar:
Thank you for your submission. While changes suggested here were never implemented, other PRs addressing the episode and challenge wording have been committed. Since this PR is stale and the issues have already been addressed, it will be closed.

PR #400, opened 2017-04-05
05: git log with 2 commits doesn't make much sense 

junghans:
git log with 2 commits doesn't make much sense. I think this is an artifact from 6e5f4aac0df66d6ad7bcca0f769917d65e622b97, which replaced "git diff --stat" with "git log --patch".


junghans:
On a second thought, we could do `git log SHA1..SHA2`.

iglpdc:
I think is simpler if you stick to your changes. Thanks @junghans!

PR #403, opened 2017-04-24
Add private email notes for GitHub

lonsbio:
Adds clarity that when using a private email address with GitHub, that address needs to be uses for user.email


iglpdc:
Thanks @lonsbio! The previous was confusing for some people at workshops. It's much clearer now.

Issue #226, opened 2016-02-03
licensing: use of the word "infective" to describe GPL


amueller:
I'm not sure if the term "infective" should be used to describe the GPL. [copyleft](https://en.wikipedia.org/wiki/Copyleft) seems to be the official term. There is some discussion here: https://en.wikipedia.org/wiki/GNU_General_Public_License#Microsoft

wking:
On Wed, Feb 03, 2016 at 11:58:39AM -0800, Andreas Mueller wrote:
> There is some discussion here:
> https://en.wikipedia.org/wiki/GNU_General_Public_License#Microsoft

Viral vs. copyleft seems to be attempts at framing the debate
depending on whether you are for or against copyleft.  Personally, I'm
for copyleft.  Given our use of command-line Git over
more-permissively-licensed replacements such as libgit2 [1] (which is
also GPLv2, it just has a linking exception), I agree that the more
GPL-friendly “copyleft” is a better term for us.

[1]: https://libgit2.github.com/

amueller:
I think it would be nice to stay neutral. I'm using permissive licenses on everything, still, I feel "infective" has stronger connotations than "copyleft". [I'm not a native speaker, though]

wking:
On Wed, Feb 03, 2016 at 01:32:38PM -0800, Andreas Mueller wrote:
> I think it would be nice to stay neutral.

Agreed.  I just don't trust myself to judge neutrality.

> I'm using permissive licenses on everything, still, I feel
> "infective" has stronger connotations than "copyleft"…

I think “copyleft” assumes familiarity with [1], although what sort of
connotations that leads is probably pretty variable ;).

http://choosealicense.com/ goes with “copyleft” and unpacks it to “I
care about sharing improvements” with a nice, cyclic-arrow icon.

[1]: https://en.wikipedia.org/wiki/Left-wing_politics

wking:
For what it's worth, choosealicense.com recently dropped it's
much-less-agressive “further restricts” phrasing [1,2].

[1]: https://github.com/github/choosealicense.com/issues/335
[2]: https://github.com/github/choosealicense.com/pull/362/commits/9ec2de1a505b8ad6dfe7c508b69ec59f6cd5103c

iglpdc:
Closing this via https://github.com/swcarpentry/git-novice/commit/bb04e50a1d33007855795e5511168aaeea9f8e30.

PR #421, opened 2017-06-29
Small clean up of glossary

aragilar:
This removes 'infective license' from the glossary, as it's not used anywhere (having been removed from the licence discussion through #226, and switches to 'changeset' over 'change set', which appears to be what both git and wikipedia use.


iglpdc:
Thanks a lot,  @aragilar!

PR #425, opened 2017-07-27
Learning objective in 04 regarding commit messages

bourque:
This is for instructor training checkout.

I thought that it may be useful to have a learning objective regarding commit messages in the Tracking Changes episode.

It looks like my editor also automatically removed trailing spaces ;)


daisieh:
This seems fine, thanks!

PR #428, opened 2017-07-31
fix typo and links

ajavedm:
There was extra 'and'  in line 72
All fixed links were going 404 (page not found).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,7 +34,7 @@ and to meet some of our community members.
 2.  If you have a [GitHub][github] account,
     or are willing to [create one][github-join],
     but do not know how to use Git,
-    you can report problems or suggest improvements by [creating an issue][issues].
+    you can report problems or suggest improvements by [creating an issue][new-issue].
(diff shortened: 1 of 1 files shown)


iglpdc:
Thanks @ajavedm! But the links are supposed to be like this. The `[issues]` and `[repo]` are shortcuts defined elsewhere, so we don't have to type the whole thing every time.

The reason you see them broken though is because they are actually not defined anywhere :).

To fix them you should edit `CONTRIBUTING.md`: all the links are at the end. Note that `[issues]` and `[repo]` have a wrong `FIXME` in there.

You can see [the same issue, now fixed, in the python lesson](https://github.com/swcarpentry/python-novice-inflammation/pull/398#issuecomment-323814675).

Could you please add another commit with a similar change? You should also undo your changes (keep the removal of the extra `and` though). To do this, the easiest is to add another commit.

ajavedm:
@iglpdc I have modified the commit.

iglpdc:
Perfect, @ajavedm! Thanks a lot.

PR #432, opened 2017-08-08
Add merge conflict avoidance strategies

atz:
- include simple preventative steps before getting to breaking up files (which doesn't actually solve anything if the same lines will be logically revised together anyway)
- separate technical and social strategies
- I understand the lesson currently de-empasizes branching, but it would be a disservice to avoid referencing it here, at a minimum, since putting work on a branch is exactly how one would implement the project management strategy of scheduling "this before that", without  blocking the developer.

Honestly, if you have conflicts with collaborators and you aren't using branching, it is the **first** thing I would recommend. I downplay it here, but the inevitability of conflicts is a prime motivation.


iglpdc:
I like it very much. Thanks again, @atz!

iglpdc:
I agree with the branching argument, but we have to live without it in this novice lesson.

Issue #370, opened 2016-12-05
Reverting a Commit in Episode 5 uses "group repository"


alex-konovalov:
"Reverting a Commit" challenge in Episode 5 precedes remotes which will be introduced in Episode 7. Therefore the text 
```
> Jennifer is collaborating on her Python script with her colleagues and
> realises her last commit to the group repository is wrong and wants to
> undo it.  Jennifer needs to undo correctly so everyone in the group
> repository gets the correct change. 
```
may be confusing since it's not clear yet what does the "group repository" mean. Also, the list of commands in this challenge does not include pushing to the remote repository (maybe it's assumed that the group repository in this challenge is on a shared filesystem?).

Perhaps this challenge could be moved to some later episode when `push` and `pull` commands are already introduced. Alternatively, it could stay here, but the use case could be changed from the group one to individual - how to revert some old commit is useful knowledge in this case anyway.

tylerkelly13:
Hi Alex,
Not sure about moving it, but I have updated the challenge to (hopefully) make it clearer.

iglpdc:
Closed by #434.

PR #434, opened 2017-08-17
issue #370 - Improved wording for 'Reverting a commit'

tylerkelly13:
Per https://github.com/swcarpentry/git-novice/issues/370, this commit attempts to improve the wording of the 'Reverting a commit' challenge. 
- Changed 'group repository' to 'project repository' to be more consistent with the preceding challenge
- Improved explanation of the difference between git checkout <commit> and git revert <commit>
- replaced 'wrong' with 'error'
- general attempt to improve challenge

### DIFF
--- a/_episodes/05-history.md
+++ b/_episodes/05-history.md
@@ -353,14 +353,14 @@ moving backward and forward in time becomes much easier.
 > ## Reverting a Commit
 >
 > Jennifer is collaborating on her Python script with her colleagues and
-> realizes her last commit to the group repository is wrong and wants to
-> undo it.  Jennifer needs to undo correctly so everyone in the group
-> repository gets the correct change.  `git revert [wrong commit ID]`
-> will make a new commit that undoes Jennifer's previous wrong
-> commit. Therefore `git revert` is different than `git checkout [commit
-> ID]` because `checkout` is for local changes not committed to the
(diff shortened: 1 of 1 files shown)


alex-konovalov:
Thank you @tylerkelly13 - I like multiple improvements of the wording used. I reread the text and still  think that this should me moved later, perhaps to "Collaborating". Even the 1st line of the challenge starts with "Jennifer is collaborating on he" Python script with her colleagues...".

tylerkelly13:
I agree it does seem a bit out of place, so if someone can figure out a better place for it we can move it. Maybe one of the maintainers might comment? Collaborating does make sense regarding the context, but reverting a commit should also be early on, given that it is 'undoing' an error and shows learners that such a mistake would not be permanent, making git less daunting (I suspect).

alex-konovalov:
@tylerkelly13 Indeed... maybe this should evolve into two portions: one in a local repository (that does not have remotes at all), covering reverting last commit and then reverting some commit from the past, and then later the same in the context of collaborative work. Hope maintainers may say a word.

kekoziar:
Per the above comment, "Within the context of this PR, I don't think this is necessary for the purposes of merging."  Merging the PR, since the broader discussion did not occur and is not necessary for mergine.

PR #440, opened 2017-10-05
corrected typo in exiting Vim section

vai2fc:
Just a quick correction to make the 'exiting Vim' section read more smoothly.


iglpdc:
Thanks @vai2fc !

PR #443, opened 2017-10-11
Tell students that git version differences are ok

GlazerMann:
If not warned, then students can get all uptight if there is a slight difference in the wording of the git output.


GlazerMann:
This doubly true, since at many locations people have to use their own computers or we have way too much paperwork

daisieh:
This seems like a helpful clarification, thanks!

PR #450, opened 2017-11-16
Make the git checkout command style consistent with rest of lesson

tmorrell:
The branch name and "-f" option haven't been introduced in this lesson.  Using HEAD works and is consistent with the rest of the lesson.


iglpdc:
Good catch! Thanks and welcome, @tmorrell !

Issue #418, opened 2017-06-02
Wrong output of some commands in episode 5.


ilectra:
In the part where `git diff HEAD~1 mars.txt` is discussed, the correct output should be (I'm copying from my local example, so authors and hashtags are different...) :

```
$ git diff HEAD~1 mars.txt
```
```
diff --git a/mars.txt b/mars.txt
index 315bf3a..93a3e13 100644
--- a/mars.txt
+++ b/mars.txt
@@ -1,2 +1,4 @@
 Cold and dry, but everything is my favorite color
 The two moons may be a problem for Wolfman
+But the Mummy will appreciate the lack of humidity
+An ill-considered change
```

and then
```
$ git diff HEAD~2 mars.txt
```
```
diff --git a/mars.txt b/mars.txt
index df0654a..93a3e13 100644
--- a/mars.txt
+++ b/mars.txt
@@ -1 +1,4 @@
 Cold and dry, but everything is my favorite color
+The two moons may be a problem for Wolfman
+But the Mummy will appreciate the lack of humidity
+An ill-considered change
```

Then
```
$ git show HEAD~2 mars.txt
```
```
commit b4dd35ec93adb5d8b6e4278fea9b23c992008209
Author: Ilektra Christidi <ilektra.christidi@ucl.ac.uk>
Date:   Fri Jun 2 09:27:10 2017 +0100

    Start notes on Mars as a base

diff --git a/mars.txt b/mars.txt
new file mode 100644
index 0000000..df0654a
--- /dev/null
+++ b/mars.txt
@@ -0,0 +1 @@
+Cold and dry, but everything is my favorite color
```

And finally (this is the tag of my first commit):
```
$ git diff b4dd35ec93adb5d8b6e4278fea9b23c992008209 mars.txt
```
```
diff --git a/mars.txt b/mars.txt
index df0654a..93a3e13 100644
--- a/mars.txt
+++ b/mars.txt
@@ -1 +1,4 @@
 Cold and dry, but everything is my favorite color
+The two moons may be a problem for Wolfman
+But the Mummy will appreciate the lack of humidity
+An ill-considered change
```

I suspect the last change (adding the line "An ill-considered change") was commited before all the `git diff HEAD....` commands, that's why the output is different from mine. But that doesn't seem it was intended.

PR #451, opened 2017-11-16
Correct `git diff` and `git show` output in Exploring History episode

mwoodbri:
As reported in #418, a previous change has led to the output of some commands presented in episode 5 of Version Control with Git becoming inconsistent with the lesson content. I have independently replicated the issue, and this PR corrects the content appropriately.


iglpdc:
Thanks, @mwoodbri !

Issue #272, opened 2016-05-20
Why are nested repos a bad idea?


wking:
This just came up in discuss@ today, and I wasn't sure.  Checking the history here, the idea [landed][1] in 0917191 (Preparing material for Version 5 novice release, 2014-02-07), but I haven't been able to find motivation.  I setup nested repos in a number of cases (e.g. building from master in the parent repo to populate gh-pages in a nested repository), so I don't see a problem with this workflow.  But I use Git every day, so maybe the problem is just that this might be confusing users who forget which repo they're in?  Or having them forget to gitignore the sub-repo in the parent-repo, and then accidentally add the sub-repo in a parent commit?

[1]: https://github.com/swcarpentry/git-novice/commit/0917191ac4a0d00f569bdb3818397f03f25ef0f8#diff-1250a30423c4241b6caf96ff324090e7R723

jttkim:
My quick answer is that Git commands find the repo meta information by starting to look for `.git` in the current working directory, and if there's none they keep going to the parent, until they either find it or they end up at the root without any luck, at which point you'll get the

    fatal: Not a git repository (or any of the parent directories): .git

message. However, if there are `.git` directories in any subdirectories, that doesn't matter, the meta information for the entire tree is held in the top-level `.git` directory. As a result, it's quite possible to create a situation where `git status` reports a file in a subdirectory as new (or modified), whereas when you run `git status` in the subdirectory which is a "nested repo", it will appear untracked. And that's just one example -- the basic issue is that a repo tracks files in all its subdirectories, so duplicating that type of info in subdirectories creates a denormalised situation which will result in inconsistencies sooner or later.

wking:
On Fri, May 20, 2016 at 01:10:32PM -0700, Jan T. Kim wrote:
> As a result, it's quite possible to create a situation where `git
> status` reports a file in a subdirectory as new (or modified),
> whereas when you run `git status` in the subdirectory which is a
> "nested repo", it will appear untracked.

Ok, so it is just an “I forgot to add a gitignore entry” problem.  I
agree that trying to track the same subdirectory in both a parent repo
and a nested repo is a bad idea, but gitignore in the parent lets you
nest repos and tell the parent not to track anything in the subrepo.

jttkim:
ok... I guess the "no nested repo" rule is one of those "as with most rules, these may be broken if you know what you're doing" [1] one.

[...]

PR #459, opened 2017-12-14
Removed advanced topics from Episode 3: Creating a Repository

AmyOlex:
Hello,

In this pull request I've made some wording changes to Episode 2, and have suggested a new wording to the key point so that it better reflects the learning objectives.  I have also made significant changes to the challenge exercise in Episode 3.  Upon working through this myself, and then reading some of the comments from Issue #272 I found this module to stray from the basic message of how to create a Git repo and where Git stores its repo data.  I think the nested repo topic is important, but needs to be distilled down for the novice.  In my proposed changes I remove the content about ignoring files and adding directories as those are topics addressed in future lessons.  I tried to focus it so that a novice would understand why nested repos were 1) not necessary, and 2) not a good idea.  Any feedback on my proposed changes is welcome.


kekoziar:
Thank you for proposing the changes.  This PR also resolves issue #476.

PR #463, opened 2018-01-09
Explicitly call module out as supplemental

jduckles:
I was just asked by an instructor if they had to teach the RStudio episode of the Git lesson, as the schedule looks now, it looks that way. I added the word "supplemental" to note that it is not a required episode when viewing the schedule.


iglpdc:
Thanks, @jduckles !

PR #469, opened 2018-01-16
Update 09-conflict.md

JamesEMcClure:
These changes clarify when a merge conflict is identified by git. In this case, the git push command fails not because of the merge conflict, but because the remote repository has newer changes that have not yet been incorporated into the local repository. Even if a merge conflict does not exist, git pull will fail in this scenario. The merge conflict is only identified after git fetch retrieves the changes made to the remote repository into the local version and then attempts to merge the two versions with git merge.  This necessitates manual intervention to complete the failed merge operation.

These changes are part of my software carpentry instructor training checkout process.

### DIFF
--- a/_episodes/09-conflict.md
+++ b/_episodes/09-conflict.md
@@ -133,8 +133,8 @@ hint: See the 'Note about fast-forwards' in 'git push --help' for details.
 
 ![The Conflicting Changes](../fig/conflict.svg)
 
-Git detects that the changes made in one copy overlap with those made in the other
-and stops us from trampling on our previous work.
+Git rejects the push because it detects that the remote repository has new updates that have not been
+incorporated into the local branch.
 What we have to do is pull the changes from GitHub,
 [merge]({{ page.root }}/reference/#merge) them into the copy we're currently working in,
 and then push that.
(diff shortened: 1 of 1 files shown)


_episodes/09-conflict.md
```diff
@@ -158,8 +158,12 @@ Automatic merge failed; fix conflicts and then commit the result.
 ~~~
 {: .output}
 
-`git pull` tells us there's a conflict,
-and marks that conflict in the affected file:
+The `git pull` command combines a `git fetch` with a `git merge` to update the local repository to include
```
iglpdc: I'm not sure whether introducing `git fetch` and `git merge` is good. People may get confused and think they miss the part where the commands where explained. What do you think about talking about "fetching" and "merging" without mentioning them as `git` commands? 

A few lines later with talk about fetching the repo and a few lines before we talk about mergin changes.

JamesEMcClure: I agree with this. I think this will be clear without the specific reference regarding git pull as a git fetch + git merge.

iglpdc: @JamesEMcClure, great! Could you make the changes in your PR? I think "git pull" to "pull" and "git merge" to "merge" will make it.

JamesEMcClure: I've updated my pull request so that the specific references to the "git fetch" and "git merge" commands no longer appear.

iglpdc:
Very nice, @JamesEMcClure ! Thanks again!

PR #472, opened 2018-01-23
03-create: Clarify Git 'repository'

mcritchlow:


iglpdc:
Thanks, @mcritchlow ! Good catch!

Issue #461, opened 2017-12-20
Lesson 1 tape analogy


anderspitman:
Comparing git to a tape might not be meaningful, especially to younger learners.

iglpdc:
Addressed by #473

PR #473, opened 2018-01-29
01-basics - Update language for clarity

caseyyoungflesh:
Changed some of the language to be more clear - changes in sentence structure, defining jargon. Also changed the use of the 'tape' analogy to 'recording' (everyone knows what a recording is!) in reference to issue #461 .

### DIFF
--- a/_episodes/01-basics.md
+++ b/_episodes/01-basics.md
@@ -30,14 +30,14 @@ history](https://support.google.com/docs/answer/190843?hl=en), or LibreOffice's 
 Version control systems start with a base version of the document and
 then record changes you make each step of the way. You can
 think of it as a recording of your progress: you can rewind to start at the base
-document, play back each change you made, eventually arriving at your
+document and play back each change you made, eventually arriving at your
 more recent version.
 
 ![Changes Are Saved Sequentially](../fig/play-changes.svg)
 
 Once you think of changes as separate from the document itself, you
-can then think about "playing back" different sets of changes, resulting in 
-different versions of the document. For example, two users can make independent
+can then think about "playing back" different sets of changes on the base document, ultimately
+resulting in different versions of that document. For example, two users can make independent
 sets of changes on the same document. 
 
 ![Different Versions Can be Saved](../fig/versions.svg)
--- a/_episodes/01-basics.md
+++ b/_episodes/01-basics.md
@@ -28,41 +28,40 @@ Word's [Track Changes](https://support.office.com/en-us/article/Track-changes-in
 history](https://support.google.com/docs/answer/190843?hl=en), or LibreOffice's [Recording and Displaying Changes](https://help.libreoffice.org/Common/Recording_and_Displaying_Changes).
 
 Version control systems start with a base version of the document and
-then save just the changes you made at each step of the way. You can
-think of it as a tape: if you rewind the tape and start at the base
-document, then you can play back each change and end up with your
-latest version.
+then record changes you make each step of the way. You can
+think of it as a recording of your progress: you can rewind to start at the base
+document, play back each change you made, eventually arriving at your
+more recent version.
 
 ![Changes Are Saved Sequentially](../fig/play-changes.svg)
 
 Once you think of changes as separate from the document itself, you
-can then think about "playing back" different sets of changes onto the
-base document and getting different versions of the document. For
-example, two users can make independent sets of changes based on the
-same document.
+can then think about "playing back" different sets of changes, resulting in 
+different versions of the document. For example, two users can make independent
+sets of changes on the same document. 
 
 ![Different Versions Can be Saved](../fig/versions.svg)
 
-Unless there are conflicts, you can even play two sets of changes onto the same base document.
+Unless there are conflicts, you can even incorporate two sets of changes into the same base document.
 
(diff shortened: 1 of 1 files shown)


_episodes/01-basics.md
```diff
@@ -28,41 +28,40 @@ Word's [Track Changes](https://support.office.com/en-us/article/Track-changes-in
 history](https://support.google.com/docs/answer/190843?hl=en), or LibreOffice's [Recording and Displaying Changes](https://help.libreoffice.org/Common/Recording_and_Displaying_Changes).
 
 Version control systems start with a base version of the document and
-then save just the changes you made at each step of the way. You can
-think of it as a tape: if you rewind the tape and start at the base
-document, then you can play back each change and end up with your
-latest version.
+then record changes you make each step of the way. You can
+think of it as a recording of your progress: you can rewind to start at the base
+document, play back each change you made, eventually arriving at your
```
somexpert: I think this is a good change, since of course tapes are much less common now unless you are serious about long-time off-site data preservation. However, I would recommend replacing the comma after "document" with the word "and".

_episodes/01-basics.md
```diff
@@ -28,41 +28,40 @@ Word's [Track Changes](https://support.office.com/en-us/article/Track-changes-in
...
 ![Changes Are Saved Sequentially](../fig/play-changes.svg)
 
 Once you think of changes as separate from the document itself, you
-can then think about "playing back" different sets of changes onto the
-base document and getting different versions of the document. For
-example, two users can make independent sets of changes based on the
-same document.
+can then think about "playing back" different sets of changes, resulting in 
+different versions of the document. For example, two users can make independent
+sets of changes on the same document.
```
somexpert: This explanation is slightly more concise, but I think removing another mention of the "base document" might be a missed opportunity to reinforce the desired mental model. I would stick that back in and keep the rest of your changes.

_episodes/01-basics.md
```diff
@@ -28,41 +28,40 @@ Word's [Track Changes](https://support.office.com/en-us/article/Track-changes-in
...
-across different computers facilitating collaboration among different people.
+across different computers, facilitating collaboration among different people.
 
 > ## The Long History of Version Control Systems
 >
 > Automated version control systems are nothing new.
 > Tools like RCS, CVS, or Subversion have been around since the early 1980s and are used by many large companies.
-> However, many of these are now becoming considered as legacy systems due to various limitations in their capabilities.
-> In particular, the more modern systems, such as Git and [Mercurial](https://swcarpentry.github.io/hg-novice/)
+> However, many of these are now considered legacy systems (i.e., outdated) due to various limitations in their capabilities.
```
somexpert: Definitely think adding this definition is a great idea. Not all our learners may know this term, particularly those who learned English as a second language. :thumbsup:

_episodes/01-basics.md
```diff
@@ -28,41 +28,40 @@ Word's [Track Changes](https://support.office.com/en-us/article/Track-changes-in
...
+> More modern systems, such as Git and [Mercurial](https://swcarpentry.github.io/hg-novice/),
 > are *distributed*, meaning that they do not need a centralized server to host the repository.
```
somexpert: Would it make sense to also briefly mention the benefit of not having a single point of failure?

somexpert: Something like: "meaning that they do not need a centralized server to host the repository, since every clone is a complete copy of the version history."?

_episodes/01-basics.md
nhejazi: Weighing in as a maintainer, this change looks good to me. I too agree with the addition of "and" after the use of "document". Maybe you could update your PR with this minor change @caseyyoungflesh?

_episodes/01-basics.md
nhejazi: Again, agree with the comment from @somexpert above. Should be a very minor change to the PR.

_episodes/01-basics.md
nhejazi: 👍

_episodes/01-basics.md
nhejazi: Just my $0.02, but I think it might be a bit too much to mention the difference between systems like SVN and Git, which is what I understand the comment above to be @somexpert. In my experience, most learners come in to this lesson not knowing what version control is at all and this is a level of detail which might not be needed.

somexpert: Yeah, I think you're right, it's not very useful information in this case.

_episodes/01-basics.md
caseyyoungflesh: Agreed. Will make that change

_episodes/01-basics.md
caseyyoungflesh: Sounds good

_episodes/01-basics.md
caseyyoungflesh: Obligatory emoji 👍

_episodes/01-basics.md
caseyyoungflesh: Would agree with @nhejazi here. KISS (Keep It Simple Stupid) at work

caseyyoungflesh:
Requested changes made!

nhejazi:
Great, thanks for the quick work on this!

PR #475, opened 2018-02-08
Fix licensing journal link

### LINKED ISSUE #474: Licensing article link is broken
https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1002598 gave me `ERR_CONNECTION_TIMED_OUT` error.

It is mentioned in https://swcarpentry.github.io/git-novice/11-licensing/

pllim:
Fix #474

Thanks for the correct link, @nhejazi !

### DIFF
--- a/_episodes/11-licensing.md
+++ b/_episodes/11-licensing.md
@@ -77,4 +77,4 @@ thus need to be asked for approval once a license is chosen.
 > - [EtherPad](https://github.com/ether/etherpad-lite), a real-time collaborative editor
 {: .challenge}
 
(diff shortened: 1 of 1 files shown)


iglpdc:
Thanks, @pllim ! But could you use this link instead:  https://doi.org/10.1371/journal.pcbi.1002598?

It redirects to the same webpage, but uses a [DOI](https://en.wikipedia.org/wiki/Digital_object_identifier), which is supposed to be persistent, and the secure version of http, [https](https://en.wikipedia.org/wiki/HTTPS), which is more secure.

pllim:
Done. Thanks!

iglpdc:
Thanks again, @pllim !

Issue #353, opened 2016-10-26
What happened to the 'bio repository' exercise


lexnederbragt:
There used to be an exercise  asking the learners to go throught the 'git init, add, commit, diff' cycle, but it is now no longer there. Is that for a particular reason? I found the text in a Socrative quiz:

Git workflow exercise

1) Create a new Git repository on your computer called bio.
2) Write a three-line biography for yourself in a file called me.txt and commit your changes.
3) modify one line, add a fourth line, and display the differences between its updated state and its original state.

kekoziar:
Exercise removed [7/17/2016](https://github.com/swcarpentry/git-novice/commit/64da97a8440f43cbedd6841ff7b13d28ee44b681#diff-3c9b1b2841b52dffd10441f1fcf62f8e); no reason found.  Closing as this doesn't apply anymore.

PR #477, opened 2018-02-13
Putting back the 'bio' exercise

lexnederbragt:
See #353


iglpdc:
Thanks! Could you add the solution, @lexnederbragt ? Most of it is already in the solution to the next exercise...

lexnederbragt:
Good suggestion. Done!

iglpdc:
Thanks a lot, @lexnederbragt ! Very nice.

PR #486, opened 2018-03-21
Rephrase to avoid "simply"

katrinleinweber:
There might even be the opportunity here to add a (possibly only slightly funny) "only look, don't touch" to this context to drive the point about Git's detached HEAD state home.


iglpdc:
Good catch, @katrinleinweber ! Do you wanna go for the "look, don't touch" or prefer to merge it as it is?

iglpdc:
Awesome, @katrinleinweber ! Thanks!

katrinleinweber:
You're welcome, and thanks for merging :-)

PR #490, opened 2018-03-27
added objective about describing collaborative workflow

shals3:
I added an additional objective which resolves Issue # 316 for the collaboration episode.


iglpdc:
Thanks and welcome, @shals3! Sounds good to me: I'm merging.

PR #498, opened 2018-04-09
Update gitignore commit message

odarbelaeze:
The message was not in agreement with the lesson thought in the "tracking changes" episode.

Hi, sorry for not creating an issue for this first, however, I noticed in a practice lesson that this commit message doesn't agree with the lessons tought before, in a simmilar exercise, this message would be categorised as "not descriptive enough"


iglpdc:
Hi, @odarbelaeze ! You're right: your version describes much better what's actually being done. Thanks for your contribution and welcome!!

Issue #360, opened 2016-11-04
using gui editors for commit messages


chendaniely:
Notes from the 2016-11-04-nsbe workshop:

students on windows machines are unable to get sublime text working from git bash
even after they copy paste the git config from the setup page in the lesson.

jsan4christ:
There is a very useful guide here how to add text editors to Git, following it I was able to atom as my default edit. Let us know if you still help.
https://help.github.com/articles/associating-text-editors-with-git/

iglpdc:
Thanks for the link @jsan4christ !

We have most of them in the episode on how to [setup git](http://swcarpentry.github.io/git-novice/02-setup/). We are missing some, though: atom for Windows and Textmate for Mac.

I didn't want to have an extesive list of editors (they come and go and it's hard to maintain), but we are pretty much there, so we could add these to complete the pack.

Would you like to send a PR with this? A link to the Github page you qouted would be helpful too (they probably would keep it very much up to date).

kekoziar:
Original issue addressed in https://github.com/swcarpentry/git-novice/pull/499
Suggestion can be searched for and revisited if someone wants to actively work on it.

PR #499, opened 2018-04-13
Ensure Sublime works without symlinking (related to #360)

katrinleinweber:
Hello :-)

The provided command did not work on my macOS 10.13.3 with Sublime 3.0. https://stackoverflow.com/a/22632139 helped and explained that for the original command to work, a symlink had to be set up. To avoid pushing that to learner, but still providing a working command, I suggest to mirror the SO solution.


katrinleinweber:
You're welcome, and thanks for merging :-)

PR #504, opened 2018-05-01
03-create.md: fix motivatingexample img path

ldko:
This should fix the relative path to the motivatingexample.png image that currently appears to result in a broken image.


iglpdc:
Thanks, @ldko ! Good catch! (And welcome!)

PR #520, opened 2018-05-13
Remove incorrect/misleading parenthetical clause

alexanderzimmerman:
When reviewing this lesson, I was surprised to see the suggestion to pronounce "HEAD~1" as "head minus one". If anything, it should just be pronounced "head tilde one". I can only imagine the other pronunciation confusing learners. I did some Googling and couldn't find anybody else calling it "head minus one".  Since this is such an easy change, I just went ahead and implemented it, rather than making an issue.

### DIFF
--- a/_episodes/05-history.md
+++ b/_episodes/05-history.md
@@ -117,7 +117,7 @@ In this way,
 we can build up a chain of commits.
 The most recent end of the chain is referred to as `HEAD`;
 we can refer to previous commits using the `~` notation,
-so `HEAD~1` (pronounced "head minus one")
+so `HEAD~1`
 means "the previous commit",
 while `HEAD~123` goes back 123 commits from where we are now.
 


iglpdc:
Thanks, @alexanderzimmerman !

Hmmm, I think we should keep some details on how to read it. Note that many learners may be unfamiliar with the name of the symbol, non-native English speakers, or following the lesson at home. I agree that probably 'pronounce as' is not the right wording, though.

What about keeping the sentence but changing 'pronounced minus' to 'pronounced tilde', so we give the correct way of reading it and also name the symbol appropiately?. I think the rest of the text as it is now is clear enough to understand that the meaning of `~` is "minus".

alexanderzimmerman:
Sounds good. I'll go ahead and make that change as soon as I find the way to type the pronunciation of "tilde" :)

iglpdc:
Yep! Thanks a lot, @alexanderzimmerman !

slel:
May I politely disagree with

> If anything, it should just be pronounced "head tilde one".

My view on this is that `HEAD~1` refers to the commit that is
one commit before head, which one can logically refer to as
"head minus one". So I don't think it's wrong to pronounce
the meaning rather than the spelling.  To address the potential
surprise at the mismatch between the two, we could improve
the phrasing to "(pronounced 'head minus one' but spelled
with a tilde sign rather than a minus sign)".

May I also politely disagree with

> Since this is such an easy change, I just went ahead and
> implemented it, rather than making an issue.

Whether or not the change is easy to make might have little
to do with whether or not we want to make this change or not.

I think issues are good for discussing whether or not we want
the change or not, whether this surprising phrasing which
looks like a mistake is really a mistake... And the pull request
discussion is good for discussing the technical details of how
the change is made.

alexanderzimmerman:
@slel, as you say, the pull request format is really nice for encouraging these discussions. I only went ahead and made the change to my own fork, and then opened the PR so that we could discuss and the powers that be (i.e. @iglpdc in this case) could make merge decisions. To me, changing my fork in this way is the clearest way to describe my intent.

iglpdc:
@alexanderzimmerman As @slel says, it's better to open an issue first. In this way, most of the discussion happens in the issue. If we decide the way to fix the issue is to make some changes, we open a PR related to the issue. The discussion on the PR is about the changes in the PR, not so about the fundamental thing to be changed (this was decided in the issue discussion).

Very small changes which need no discussion could skip the issue part (say a typo). This is somehow in between.

Going back to the issue at stake, I think part of the problem was the use of the word "pronounce". I think there is value in stating the name of the symbol as tilde (as I said before, many learners have not used the symbol before) and its meaning in this context as "minus". Maybe something like '`~` read as "tilde" and meaning "minus"' or the wording @slel suggested above, is better.

There is also a weird thing happening in workshops. If the instructor reads `~` as "minus" many learners would type `-` because it's hard to see the difference for those sitting back in the room. So I think saying aloud `~` as tilde, at least once, is valuable.

alexanderzimmerman:
@iglpdc and @slel , if you feel strongly about encouraging this procedure for issues and pull requests, then I think someone should open an issue to fix [https://github.com/swcarpentry/git-novice/blob/gh-pages/CONTRIBUTING.md](the contribution instructions) which I read at the beginning of my instructor checkout process.

As for the rest of the discussion, as you suggest, an issue should be raised (referencing this PR) rather than continuing the discussion in this PR.

Issue #522, opened 2018-05-13
Staging process chart used to show "state"


agzimmerman:
[This lesson](https://swcarpentry.github.io/git-novice/04-changes/index.html) introduces a nice chart showing the staging and committing process:
![image](https://user-images.githubusercontent.com/9367569/39967081-fa35625e-56b5-11e8-9d4b-868ba9af70a8.png)

But then in [a later lesson](https://swcarpentry.github.io/git-novice/07-github/index.html), a similar chart is used to show the "state" of a repo:
![image](https://user-images.githubusercontent.com/9367569/39967095-27017944-56b6-11e8-80e1-e50a6e05e840.png) 
This is done again in [the next lesson](https://swcarpentry.github.io/git-novice/08-collab/index.html).

I fail to understand how this chart explains the state of a repository, and I don't see where this has been explained. Why do we have all of this information about the staging process in these places? I'm personally confused and distracted by it, and I imagine the learner might be as well. Is it just me?

alexanderzimmerman:
It seems we agree about not wanting to add extra load; but we have different perspectives on what adds cognitive load here.

I would argue that this graphic shows too much extraneous information. I wouldn't assume that the learner has mastered the staging area concept and is perfectly comfortable with this staging area chart. 

So for this lesson, I think we would just have a picture of a .git repo box with some files in it which represents the "state", since we don't care about the state of the staging area.

Since I'm still learning plenty of this as well: Is it true that the remote repo doesn't have a staging area, or do we just not interact with it? I think this would contradict what we teach earlier in the episode:
![image](https://user-images.githubusercontent.com/9367569/40065032-0dae8a2a-5861-11e8-9d7a-eb6a953e8d5a.png)

Thanks for thinking through this with me; and I apologize for the barrage of issues and comments that have come from me digging into this lesson for the first time :)

alexanderzimmerman:
Also, about side-effects of removing the staging area from the graphics in this episode: I think it would be quite clear if we just labeled the boxes "local" and "remote", and perhaps also keep the local path and web address (already in the graphic) next to that.

munkm:
I agree that the local path and remote URL should be kept to show the address of each repo on the systems that students might interact with. 

I also agree that adding commit IDs to the stack would be a helpful addition to the graphic. When I teach the .git lesson, I generally do this on a whiteboard and I find it helpful. 

[...]

PR #527, opened 2018-05-16
ENH: Repeat local repo diagram in github lesson

aragilar:
Includes the repository diagram from the add/commit lesson in the github lesson, to clarify the difference between the repositories.

Based on comments from #522.


munkm:
Hi @aragilar, I like your change of adding the original repo from the earlier lesson. This helps to make things clearer. Thanks for your contribution! 🎉

PR #528, opened 2018-05-16
Update 04-changes.md

kyclark:
Sorry I forgot to make a commit message (esp when talking about commit messages -- doh!), but it's factually incorrect to say that one must "git add" a file in order for a commit to be accepted. One need only "add" a file once, then "commit" multiple times. The reason no commits were recorded is that the file to commit was not indicated to git. I have corrected the example and also noted the "-a" flag that says to apply the commit message to "all" files that changed.

### DIFF
--- a/_episodes/04-changes.md
+++ b/_episodes/04-changes.md
@@ -256,9 +256,8 @@ no changes added to commit (use "git add" and/or "git commit -a")
 The last line is the key phrase:
 "no changes added to commit".
 We have changed this file,
-but we haven't told Git we will want to save those changes
-(which we do with `git add`)
-nor have we saved them (which we do with `git commit`).
+but we haven't told Git we will want to save those changes 
+(which we do with `git commit`).
 So let's do that now. It is good practice to always review
 our changes before saving them. We do this using `git diff`.
 This shows us the differences between the current state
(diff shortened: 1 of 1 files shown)


iglpdc:
Thanks, @kyclark ! I'm a bit confused by your changes. I think it's correct to say that changes in a file should be added (or staged) before being commited (see docs for [git-commit](https://git-scm.com/docs/git-commit) and [git-add](https://git-scm.com/docs/git-add)).

The typical workflow (and the one we recommend in the lesson) is using `git add` to collect all the changes in several files followed by `git commit` with no file arguments. I think other alternatives, such as `git commit -a` should be used only in specific cases and once one has some experience with Git.

As you mentioned, Git supports the "Subversion-way" of commiting files in one step with `git commit -- file1 file2...`, but I think it's very error-prone (it's easy to forget a file or add unwanted changes) and should not be recommended to a new user.

kyclark:
OK, I just read the official "git-add" docs, and it agrees with your workflow. In my many years of using Git, I've never used "add" except to specifically start tracking a file, and I always "commit" specific file(s) or use "-a" if I've made a batch of changes around the same idea.

mzedevworks:
I have also been using "git commit -a" for many years but usually I have realized that it brings up two issues which are easily avoided by using "git add" first.

- Usually the commit message is not very specific as its one commit message for a lot of files and becomes misleading for others when they go through the commit messages to find out what was changed.
- The other biggest problem I found with it is having to commit changes that are not ready to be committed as this will commit everything. This gets reduced by doing  "git add" first as it will allow you to first stage only the files you want to go with this commit you are preparing.

kekoziar:
Since all posters on this PR (including the original submitter) agree `git commit -a` should not be taught to novices, I'm closing this PR.

Issue #507, opened 2018-05-02
Lesson 14 - No guidance for RStudio users when Git.exe path not pre-filled


richmccue:
One thing I ran into while trying to complete "Lesson 14 - Supplemental: Using Git from RStudio"  (version 1.1.383) on a Windows 10 computer was that the Git executable path was not pre-filled for me.

I had to go looking for the git.exe file which was buried pretty deep in the file system "Tools" -> "Global options" -> "Git/SVN": 
C:/Users/rmccue/AppData/Local/GitHubDesktop/app-1.1.1/resources/app/git/cmd/git.exe 
![git-r-integratin](https://user-images.githubusercontent.com/581697/39550467-3becd1b2-4e16-11e8-8b39-c4eddbd94b60.png)

I'm not sure my problem was due to the configuration of my Windows 10 workstation (I didn't have any problems on my Mac laptop), but if not, it might be helpful to include instructions for Windows users on how to find their git.exe file as a linked resource for learners who get stuck. I'd be happy to work on those instructions if the community would find it useful/helpful.

Cheers,

Rich

iglpdc:
Thanks, @richmccue ! I think it will be useful to have this included in the lesson, so people can solve the problem when follow the lesson at home.

Woudl you like to add the description of the fix (with the screenshot, if possible) to the lesson? As it seems to be something specific to Windows, you could do a callout called "Set the Git executable in Windows" or something like this.

bencomp:
Was this issue fixed by #531 ?

PR #531, opened 2018-05-25
Issue #507 - Episode 14 - Added guidance for RStudio users when Git.exe path not pre-filled

richmccue:
In response to Issue #507 [](https://github.com/swcarpentry/git-novice/issues/507) I added some text to Episode 14 to give some direction to users when the Git executable path in RStudio is not pre-filled for them.  This seems to be more of a problem on Windows computers.

### DIFF
--- a/_episodes/14-supplemental-rstudio.md
+++ b/_episodes/14-supplemental-rstudio.md
@@ -37,9 +37,22 @@ our computer, we choose the option "Existing Directory":
 > option on this menu. That is what you would click on if you wanted to
 > create a project on your computer by cloning a repository from GitHub.
 > If that option is not present, it probably means that RStudio doesn't know
-> where your Git executable is. See
-> [this page](https://stat545-ubc.github.io/git03_rstudio-meet-git.html)
-> for some debugging advice. Even if you have Git installed, you may need
+> where your Git executable is, and you won't be able to progress further
+> in this lesson until you tell RStudio where it is.  
+> # Find your Git Executiable
+> First make sure that you have installed either [Git](https://git-scm.com/downloads/)
+> or [GitHub](https://desktop.github.com/) on your computer.
+> Next, open your shell on Mac or Linux, or on Windows open the command prompt.
+> and then type:
+> - 'which git' (Mac, Linux)
+> - 'where git' (Windows)
+> Copy the path to the git executable (e.g. On a Windows computer the path was:
+> 'C:/Users/UserName/AppData/Local/GitHubDesktop/app-1.1.1/resources/app/git/cmd/git.exe')
+> # Tell RStudio where to find GitHub
+> In RStudio, go to the 'Tools' menu > 'Global Options' > 'Git/SVN' and then
+> browse to the git executable you found in the command prompt or shell. Now restart
+> RStudio.
+> Note: Even if you have Git installed, you may need
 > to accept the XCode license if you are using macOS.
 {: .callout}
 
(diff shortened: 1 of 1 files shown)


murraycadzow:
This looks like a really good addition.
It might be worth rephrasing the sentence about making sure that git is installed. A learner might have git installed as part of the OS and it's not being found and didn't actually download it themselves as part of the setup. Maybe have the urls as a suggestion if it isn't installed.

munkm:
For transparency: I requested review from some of the maintainers from the r lessons that regularly use git with RStudio to weigh in on this PR. 

This PR looks great to me @richmccue . Would you mind updating your PR with @murraycadzow 's suggestion and then I'll merge it into the lesson. The clarifying work you've done here is really helpful and I think it will be appreciated by all of the learners interested in git+RStudio.

diyadas:
Looks good to me!

katrinleinweber:
Due to https://github.com/carpentries/workshop-template/pull/574 I have a fresh macOS available that hasn't seen RStudio now. My rapproving review was from before that. Shall I test the instructions?

munkm:
Sure that sounds good @katrinleinweber.

richmccue:
I've updated the language around making sure that Git is installed as suggested by @murraycadozow (Issue #507) to take into account the fact that Git may be installed as part of the OS.

munkm:
Awesome! Thank you so much @richmccue. I'm happy to merge, but I'd like to wait until @katrinleinweber verifies that they can follow the instructions on their machine to double check everything works. 🙂

_episodes/14-supplemental-rstudio.md
```diff
@@ -40,16 +40,21 @@ our computer, we choose the option "Existing Directory":
...
-> Next, open your shell on Mac or Linux, or on Windows open the command prompt.
+> First let's make sure that Git is installed on your computer.
+> Open your shell on Mac or Linux, or on Windows open the command prompt
 > and then type:
 > - `which git` (Mac, Linux)
 > - `where git` (Windows)
-> Copy the path to the git executable (e.g. On one Windows computer which had
-> GitHub Desktop installed on it, the path was:
+>
+> If there is no version of Git on your computer, please either install [Git](https://git-scm.com/downloads/)
```
katrinleinweber: How about referring to https://carpentries.github.io/workshop-template/#git here so that we do not need to maintain extra instructions here at all? https://swcarpentry.github.io/git-novice/setup.html is already doing that as well.

katrinleinweber: Which [can not be recovered](https://en.wikipedia.org/wiki/Sunk_cost) unfortunately, and therefore should not be part of the decision when what gets merged ;-)

I don't mean this harshly at all; Please don't get me wrong! I'm grateful to @richmccue for alerting us to the  problem and for providing a fix :-) Just pointing out that previous maintainers choose to refer to the setup info that's maintained centrally.

Will test this on the weekend and report back.

munkm: Thank you for the helpful wikipedia link. 

I agree that this link should point to the same instructions as the setup to be consistent with the rest of the lesson.  I do not think that that merging this PR should be blocked by this link being inconsistent.

katrinleinweber:
Reading from the top again, I can't help after all. This "seems to be more of a problem on Windows computers." Presumably, "learner might have git installed as part of the OS" prompted me to to comment upon reading it initially, sorry.

munkm:
Thank you so much for all of your hard work @richmccue!

Issue #539, opened 2018-06-25
css doesn't load on setup page


maneesha:
From the lesson site (http://swcarpentry.github.io/git-novice/) if you click on "Setup" it takes you to a page where css does not load.

PR #540, opened 2018-06-25
make css load

maneesha:
Following example by @fmichonneau when css didn't load in setup page in issue #539, I'm guessing this is how to fix this here.


fmichonneau:
thanks @maneesha !

PR #542, opened 2018-07-22
DOC: Clarify nested nature of git init & separation from directory creation

### LINKED ISSUE #526: Lesson 3 - Expanded explanation of git init
As one of the key points of the lesson is that git init initializes a repository, a second line of notes could be added to explain the git init command in more detail, specifically that the repo or “storage area” being initialized includes everything in the repository, i.e all files, sub-directories, and their files.  This could help prepare learners for the “Places to Create Git Repos” activity as well. 

While teaching the git init command, I find it helpful to 1) reiterate that the creation of the directory and the creation of the repo are two different and independent processes and 2) that initializing the repo includes everything included in it.   This is not explicitly stated, but maybe it should be?

tylerjereddy:
Fixes #526.

As discussed in the linked issue, clearly explaining that `git init` includes subdirectories is useful, especially given that the episode allocates 0 minutes to the exercises that explore this.

I've also reiterated that directory creation and repository initialization are completely separate processes, which is also considered a useful point in the linked issue.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -37,6 +37,13 @@ $ git init
 ~~~
 {: .bash}
 
+It is important to note that `git init` will create a repository that
+includes subdirectories and their files---there is no need to create
+separate repositories nested within the `planets` repository, whether
+subdirectories are present from the beginning or added later. Also, note
+that the creation of the `planets` directory and its initialization as a
+repository are completely separate processes.
+
 If we use `ls` to show the directory's contents,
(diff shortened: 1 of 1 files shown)


nhejazi:
Having just taught this lesson again recently, I think explicitly clarifying this would have been quite useful to first-time learners.

munkm:
Yep! I totally agree that this resolves #526! Thank you for your contribution @tylerjereddy!

PR #546, opened 2018-08-07
Update 02-setup.md

grantsayer:
Added the options placeholder for the command git verb options to outline that apart from the verb there can be additionally extra information for the action/verb.


munkm:
looks good to me. This is a good clarification. Thanks for for your contribution!

Issue #406, opened 2017-05-07
Using ungit or gitless as alternatives to git shell


Seanny123:
Somewhat related to the discussion in #340 has anyone tried using [ungit](https://github.com/FredrikNoren/ungit)  and/or [gitless](gitless.com) instead of a pure git shell. 

Ungit is GUI for understanding git. 
Gitless is the output of a research project trying to simplify git.

I believe they both meet the requirements of Software Carpentry given they are:

- Open Source
- Cross platform

unode:
Hi Sean,

I have tried gitless, saw a few videos on ungit and have tried many different GUIs..

On **gitless**:

I very much enjoyed the attempt of sanitizing git commands in gitless. There are more commands than with basic git but overall it makes for a more intuitive set of actions. The developers also decided to hide the staging complexity. As staging is one of the concepts that is sometimes hard to understand by learners, this helps in making it a little bit more beginner friendly.

On the other hand, roughly 60-70% of the commands do the same actions so you still have to learn those concepts. In this regard I think it adds little to bare git.

Finally, there is one big issue, gitless is not available for Microsoft platforms, only MacOS and Linux.

On **ungit**:

The graphical interface is helpful at having a visual representation of the commit history. It also provides mouse actions (click, drag&drop) to most commands.

While this can be helpful to avoid the command-line I think it brings little didactic value. Similarly to gitless you still need to have an understanding of the concepts behind git.

To some extent I also think that a graphical interface can be counter-productive since the learner can be distracted by eye-candy elements and unnecessary complexity (a graph of a busy repository can be quite scary).

Finally, while being open-source and cross-platform, installation is more complicated since you require nodejs, npm and git beforehand.

**Overall**, while these most tools aim at making git _easier_ to use, I personally don't think they succeed.
From my experience most of the difficulty with using git comes from understanding the concepts behind it and not necessarily the commands you need to type or the buttons you need to click.

Seanny123:
Thank you for taking the time to reply @unode. Given your explanation, I now understand how these approaches don't solve the problem with understanding git.

PR #550, opened 2018-09-06
Added a new lesson on GitHub Desktop.

gauthambrahma:
This lesson primarily targets researchers who are not very found of the terminal to adopt version control using git. It can be a good replacement for some of the lessons after "Conflicts" which are non hands-on.


rgaiacs:
I think that have a episode about GitHub Desktop goes on the opposite direction of some of our values, specially be inclusive to users of different OS. https://desktop.github.com/ says

> Download for macOS or Windows (msi)

so users of Linux would not be able to follow this lesson. If we **really** want to have a lesson using a GUI for Git I would suggest to write a separate lesson, e.g. https://github.com/swcarpentry/git-novice-gui, and use https://www.gitkraken.com/ or another GUI that supports macOS, Window and Linux. For reference, [Git Bash includes an Tcl/Tk based graphical user interface](https://git-scm.com/docs/git-gui) that is part of Git so is available for macOS and Linux.

_includes/all_figures.html
```diff
@@ -53,3 +53,15 @@
 <p><img alt="GitHub Repository After First Push" src="../fig/github-repo-after-first-push.svg" /></p>
 <hr/>
 <p><img alt="The Conflicting Changes" src="../fig/conflict.svg" /></p>
+<hr/>
```
tkphd: The `</p>` tags are highlighted in red because they are within the incomplete `` tags.
Please replace `</p>` with ` /></p>`.

gauthambrahma: Thank you.

munkm:
Thank you for your pull request @gauthambrahma! You've done a lot of work and it looks like you've really thought about the content you'd like to see added.

I agree with @rgaiacs that we want to make sure that we're inclusive to users with different operating systems. I'd like to mention that there have been other issues in this repository that talk about using GUI-based alternatives in this lesson, e.g. swcarpentry/git-novice#406 and swcarpentry/git-novice#340 

I'd also like to point out that the  [contributing guidelines](https://github.com/swcarpentry/git-novice/blob/gh-pages/CONTRIBUTING.md) for this lesson cautions adding completely new lesson material, i.e. :

> Our lessons already contain more material than we can cover in a typical workshop, so we are usually not looking for more concepts or tools to add to them. As a rule, if you want to introduce a new idea, you must (a) estimate how long it will take to teach and (b) explain what you would take out to make room for it. The first encourages contributors to be honest about requirements; the second, to think hard about priorities.

The git lesson certainly tries to cover a lot, and we're intentionally leaving out other important concepts like [branching](https://github.com/swcarpentry/git-novice/pull/303) because the lesson is too long already. What would you remove from this lesson to add your new content? Is it worth prioritizing over the content that you're suggesting get deleted? As much as teaching git via a GUI could be really useful, I wonder if is in scope for this particular lesson. I would also second @rgaiacs suggestion that a gui-based lesson could be offered as an alternative to this lesson, so the entire course could be taught with the gui rather than switching from the command-line to the GUI halfway through.  Switching midway through the lesson or showing a GUI option at the end might be even more confusing to new learners.

munkm:
Hi @gauthambrahma. I'm going to close this PR because the concept isn't one we're going to be able to merge into this lesson. Please do keep us updated if you decide to make a variant of the git lesson with github desktop though. 🖥 We'd really like to hear how it goes!

PR #554, opened 2018-09-13
Wording on core.autocrlf explanation

BTopcuoglu:
"For reasons that are long to explain," could be seen as discouraging language by a novice git user. We provide them with the link to the Github page where they can learn more about line endings. I think even if they are curious to learn, they might be reluctant to open up the Github page if they worry how long it might take.

### DIFF
--- a/_episodes/02-setup.md
+++ b/_episodes/02-setup.md
@@ -44,7 +44,10 @@ in a later lesson will include this information.
 > Different operating systems use different character(s) to represent the end of a line.
 > (You may also hear these referred to as newlines or line breaks.)
 > Because Git uses these characters to compare files,
-> it may cause unexpected issues when editing a file on different machines.
+> it may cause unexpected issues when editing a file on different machines. 
+> Don't hesitate to read more about this issue on your own time
+> [on this GitHub page](https://help.github.com/articles/dealing-with-line-endings/).
+{: .callout}
 >
 > You can change the way Git recognizes and encodes line endings
 > using the `core.autocrlf` command to `git config`.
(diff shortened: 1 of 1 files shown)


munkm:
This is a good point. We want to be encouraging to learners that might be curious about this. I do think a novice instructor might get caught up here if we leave this without some sort of caveat, though. @BTopcuoglu would you mind adding an additional sentence to your PR that encourages motivated learners to learn why, but that it is beyond the scope of the git lesson.

_episodes/02-setup.md
```diff
@@ -41,11 +41,13 @@ in a later lesson will include this information.
...
 > As with other keys, when you hit <kbd>Return</kbd> on your keyboard,
 > your computer encodes this input as a character.
-> For reasons that are long to explain, different operating systems
-> use different character(s) to represent the end of a line.
+> Different operating systems use different character(s) to represent the end of a line.
 > (You may also hear these referred to as newlines or line breaks.)
 > Because Git uses these characters to compare files,
-> it may cause unexpected issues when editing a file on different machines.
+> it may cause unexpected issues when editing a file on different machines. 
+> Don't hesitate to read more about this issue on your own time
```
tkphd: "You can read more about this issue" is a cleaner statement than "Don't hesitate to read more about this issue on your own time." Recommend restoring original phrasing, but moving to the new location (line 49 rather than line 68).

BTopcuoglu: @tkphd I agree with your review. I'll restore the original phrasing "You can read more about this issue. What about the end where I've added "on your own time" to indicate to the instructor to move on?

munkm: Might I suggest something like "Though it is beyond the scope of this lesson, you can read more about this issue ...". That way it's clear to new instructors to not get tripped up explaining this in the lesson.

munkm: And yes, I agree with adding the "on your own time" to the end. It's clarifying and helpful.

BTopcuoglu: Oh that's good! Thanks @munkm

munkm:
Awesome PR @BTopcuoglu 🎉 🎈  🎆 !! Thanks for being so responsive and updating your PR. I still get tripped up teaching this sometimes, so I really like your more encouraging language.

munkm:
That's really exciting @BTopcuoglu. I hope you have a great time teaching it! Good luck!

_episodes/02-setup.md
tkphd: The new phrasing is much better. Thanks @BTopcuoglu!

PR #555, opened 2018-09-13
Fix output of git status after checking out file

jaanos:
I have noticed that the output of `git status` after a file had been checked out from a previous commit is wrong, so here's a fix. The text already correctly stated that the file would be staged.


kekoziar:
Looks good to me.  Thanks for the contribution!

Issue #558, opened 2018-09-24
Lesson 3: Add description for force delete in Solutions


krishnanj:
In [[lesson-3/ Creating a Repository]](http://swcarpentry.github.io/git-novice/03-create/index.html) of Version Control with Git there is a solution at the end which describes how to remove the `.git` folder in the moons subdirectory.

Here the idea of recursive force deletion of folders is motivated using the command `rm -rf moons/.git`. However, from my skimming, there is no description of what is force delete or why we should use it here, only a warning to use it carefully. I think it is useful to have a longer description in the solutions section. If this issue is approved, I can submit a PR for the same.

On a similar note, I also noticed that there isn't a focus on removing files and directories with git  [correct me if I am wrong]. Perhaps its good to integrate a short description on `git rm` commands in one of the tutorials?

nhejazi:
Thanks for bringing this up; I think this is a good point. In my teaching experience, usually the Git lesson is taught after a lesson that introduces shell commands, so this lack of explanation of `rm -rf` is possibly an artifact of relying on teaching in that order. I do think it's worth adding a short description of `rm -rf` as done in #560 so that the inherent danger in possibly abusing this command is less nebulous.

On another note, this is a fairly introductory lesson and I think `git rm` and related commands might be out of scope for this level of introduction. It would be nice to see a lesson covering some of these concepts but it might fall in an appendix section of the lesson or under advanced topics.

nhejazi:
Resolved by #560.

PR #560, opened 2018-09-28
 Description for force delete

krishnanj:
This is a follow-up PR submitted with regard to the [issue](https://github.com/swcarpentry/git-novice/issues/558) on description for force delete in[ Solutions in Lesson-3/ Creating a Repository](http://swcarpentry.github.io/git-novice/03-create/index.html).

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -143,6 +143,16 @@ wording of the output might be slightly different.
 >
 > > ## Solution -- USE WITH CAUTION!
 > >
+> > ### Background
+> > Removing files from a git repository needs to be done with caution. To remove files from the working tree and not from your working directory, use
+> > ~~~
+> > $ rm filename
+> > {: .language-bash}
+> > 
+> > The file being removed has to be in sync with the branch head with no updates. If there are updates, the file can be removed by force by using the `-f` option. Similarly a directory can be removed from git using `rm -r dirname` or `rm -rf dirname`.
+> >
+> > ### Solution
+> > Git keeps all of its files in the `.git` directory.
 > > To recover from this little mistake, Dracula can just remove the `.git`
 > > folder in the moons subdirectory by running the following command from inside the `planets` directory:
 > >


nhejazi:
As noted in #558, I think this is an important contribution and should be merged since it makes the use of `rm -rf` less nebulous. There is an assumption made in the lesson about the learner's familiarity with shell commands, possibly an artifact of the fact that the Git lesson is very often taught after a shell lesson; however, I think a brief description of `rm -rf` is valuable, whether as a review or for those not necessarily familiar with shell commands.

This PR looks good to me. Thanks for the contribution!

_episodes/03-create.md
```diff
@@ -143,6 +143,16 @@ wording of the output might be slightly different.
 >
 > > ## Solution -- USE WITH CAUTION!
 > >
+> > ### Background
+> > Removing files from a git repository needs to be done with caution. To remove files from the working tree and not from your working directory, use
```
katrinleinweber: In addition to #588, I think this line is very confusing. Do you mean the staging area / index, where you write "working tree"?

_episodes/03-create.md
```diff
@@ -143,6 +143,16 @@ wording of the output might be slightly different.
...
+> > ~~~
+> > $ rm filename
+> > {: .language-bash}
+> > 
+> > The file being removed has to be in sync with the branch head with no updates. If there are updates, the file can be removed by force by using the `-f` option. Similarly a directory can be removed from git using `rm -r dirname` or `rm -rf dirname`.
```
katrinleinweber: This could also be addressed in #588: Are you talking about `git rm` here? I'm very confused by this explanation, because neither `rm` nor `git rm` actually removes a file from the repository (sticking to the terminology in https://git-scm.com/about/staging-area).

katrinleinweber:
@nhejazi @krishnanj @iglpdc: Since `rm filename` _will_ delete a file from the "working directory", I'm afraid that this PR introduced a factual error. If #588 wasn't a suitable correction, please send a PR with an alternative correction, or reversion of a6693a5.

PR #570, opened 2018-10-25
Change pager keys to lowercase

M4rkD:
Change N to n:
N searches in reverse when using less, this brings up the confusing message that no matches are found (unless the user has already previously pressed n). The key for searching forward is lowercase n.
Change Q to q:
Both Q and q are valid choices for exiting the pager. Lowercase q requires fewer keystrokes, and therefore is (arguably) simpler and a better habit to form (given a choice).


munkm:
These are great points @M4rkD . Thank you for your explanation for the changes, too! This was so thorough and really appreciated. 

And thanks for your first contribution to git-novice!

PR #571, opened 2018-11-02
WIP: Remove .gitkeep files and mention self-explanatory alternative

katrinleinweber:
Hi! Please see https://github.com/carpentries/lesson-example/issues/234 for background on this suggestion. Cheers :-)

### DIFF
--- a/_episodes/04-changes.md
+++ b/_episodes/04-changes.md
@@ -590,9 +590,10 @@ Date:   Thu Aug 22 09:51:46 2013 -0400
 >
 >    Note, our newly created empty directory `directory` does not appear in
 >    the list of untracked files even if we explicitly add it (_via_ `git add`) to our
->    repository. This is due to Git only operating on files, and is sometimes worked-
->    around by creating empty `.gitkeep` files if an otherwise empty folder is deemed 
->    to be tracked by Git. However, it's better to avoid this and look for ways to 
+>    repository. This is the reason why you will sometimes see `.gitkeep` files
+>    in otherwise empty directories. Unlike `.gitignore`, these files are not special
+>    and their sole purpose is to populate a directory so that Git adds it to
+>    the repository. However, it's better to avoid this and look for ways to 
 >    automatically create such folders when your code needs them, or try to very 
 >    briefly explain the need for the empty folder with a `.gitkeep-because-...` file.
 >
--- a/_episodes/04-changes.md
+++ b/_episodes/04-changes.md
@@ -590,10 +590,9 @@ Date:   Thu Aug 22 09:51:46 2013 -0400
 >
 >    Note, our newly created empty directory `directory` does not appear in
 >    the list of untracked files even if we explicitly add it (_via_ `git add`) to our
->    repository. This is the reason why you will sometimes see `.gitkeep` files
->    in otherwise empty directories. Unlike `.gitignore`, these files are not special
->    and their sole purpose is to populate a directory so that Git adds it to
->    the repository. In fact, you can name such files anything you like.
+>    repository. You can also use a `ReadMe.md` with a short explanation of why
+>    the otherwise empty folder needs to be tracked by Git. This has the added benefit
+>    of GitHub rendering that explanation when someone browses to the folder.
 >
 > 2. If you create a directory in your Git repository and populate it with files,
 >    you can add all files in the directory at once by:
--- a/_episodes/04-changes.md
+++ b/_episodes/04-changes.md
@@ -594,7 +594,7 @@ Date:   Thu Aug 22 09:51:46 2013 -0400
 >    around by creating empty `.gitkeep` files if an otherwise empty folder is deemed 
 >    to be tracked by Git. However, it's better to avoid this and look for ways to 
 >    automatically create such folders when your code needs them, or try to very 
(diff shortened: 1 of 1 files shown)


steve855:
I took a look at your changes and I think you have proposed an interesting idea. Making a README.md file does have a lot of advantages over the .gitkeep and I may start doing this myself. However, I don't think this is a good idea to incorporate into the lesson since lots of people still use .gitkeep files to keep empty folders under version control and if this line is  removed from the lesson, the learners may not know what they are seeing when they encounter them in the wild. I think you have a good idea but this lesson may not be the right place for it. Also, I don't know if this is a real convention or just my habit but have only seen readme and README, not  the camel case ReadMe.

I think it may be too much to add to the lesson but here are some reasons I have used .gitkeep

1. When starting a project, so I can have the full directory structure of the code fixed before the code has been written to populate it
2. When something is going to be machine generated and I want to guarantee the folder will be there, i.e. my tests sometimes generate figures to include in the documentation. I have an empty _autodata folder in my git repo so that all my tests and documentation can be written to access it and I don't need to write code to test if the folder exists into every test (the tests can execute in arbitrary order)

Those are weird reasons that go beyond the scope of the lesson but they are just to show that there are some good reasons for .gitkeep files

nhejazi:
I think perhaps others should weigh in here but I do still see value in the use of `.gitkeep` files. The question of whether to use `.gitkeep` vs `README.md` (or `.txt` or w/e) comes down to the type of project as well as specifics on packages you might be relying on. `README.md` gets rendered nicely for someone browsing a project on GitHub, but it requires that the project developer author a file that explains exactly what the purpose of a given directory is, which adds additional overhead. For example, it's self-explanatory what a subdirectory `figs/` will be used for, without the need to write a new `README.md` for the subdirectory (instead, the directory gets tracked by Git after a simple `touch figs/.gitkeep`). Also, as mentioned in comments in the linked issue, the use of `README.md` might cause problems with systems like Jekyll. In any case, I think it's worth possibly mentioning that creating a `README.md` is a more informative alternative to the use of `.gitkeep` but `.gitkeep` itself should still be mentioned as it was prior to the changes proposed in this PR.

katrinleinweber:
Thanks for your replies! Yeah, the file should maybe named `README-something.md` in order to not conflict with the main `README.md` when copied into `_site/` during a `make` procedure.

> [...] directory structure of the code fixed before the code has been written to populate it.

I would argue that code which can take care of creating any folders it needs is better, because it's not coupled to the external factor of said file.

> [...] I don't need to write code to test if the folder exists into every test

Sure, we want to be lazy, but coupling the test code to an external factor that is not really needed in the repo is a bit too lazy IMHO.

Identifying the functions that OTOH create files, but not the needed directory should rather be improved than worked around.

_episodes/04-changes.md
```diff
@@ -590,10 +590,11 @@ Date:   Thu Aug 22 09:51:46 2013 -0400
...
 >    the list of untracked files even if we explicitly add it (_via_ `git add`) to our
->    repository. This is the reason why you will sometimes see `.gitkeep` files
->    in otherwise empty directories. Unlike `.gitignore`, these files are not special
->    and their sole purpose is to populate a directory so that Git adds it to
->    the repository. In fact, you can name such files anything you like.
+>    repository. This is due to Git only operating on files, and is sometimes worked-
+>    around by creating empty `.gitkeep` files if an otherwise empty folder is deemed 
+>    to be tracked by Git. However, it's better to avoid this and look for ways to 
+>    automatically create such folders when your code needs them, or try to very 
+>    briefly explain the need for the empty folder with a `.keep-because-...` file.
```
munkm: I understand your intention here by explaining the reason for `.keep-because-` to make the reason for having the .gitkeep file obvious. But in the wild I haven't seen any projects do this. Also `.gitkeep` is complimentary in naming to `.gitignore`. I disagree with incorporating the change as it is worded here. The original text says you can name them whatever you'd like without telling learners what they should be doing.

kekoziar: @katrinleinweber, will you clarify
> However, it's better to avoid this and look for ways to automatically create such
>  folders when your code needs them.

?

kekoziar (changes requested):
I'm not sure we need the `.gitkeep` files, especially since the style is updated through the styles lesson.

scottgruber:
As mentioned in this thread knowing what a .gitkeep file is will help novices understand in case they run across it in a project using version control. I’m a new maintainer to this repository and trying to close issues and pull requests to clear the board. If there is any further comment please reopen.

katrinleinweber:
Hi @scottgruber :wave: The proposed wording change would IMHO help to convey that knowledge. Please also note that [Git's official docs don't mention `gitkeep` at all](https://git-scm.com/search/results?search=gitkeep). Thus, some more guidance than the current, neutral statement of a fact seems sensible for a _lesson_.

I've replied to the remaining review comment above. Please reopen this and evaluate the PR's content.

_episodes/04-changes.md
katrinleinweber: @kekoziar & @scottgruber: How about incorporating [my clarification above](https://github.com/swcarpentry/git-novice/pull/571#issuecomment-452216390) in some fashion?

> For example, functions that create files but not the directory should be identified and improved, rather than pre-populating repositories with empty folders.

PR #573, opened 2018-11-05
Highlight clone destination of Owner's repository

spco:
Very often my learners cause problems by not fully reading the instructions around cloning the Owner's repo. Several times I've had learners overwrite their own local repository by cloning another's into the same location. This edit highlights to learners that they should avoid this.


pagrubel:
You point is good. However, first you should say "into a different location than her own" (not to her own). Since you are trying to be explicit you might even want to say different than the existing planets directory. Instead let's designate a directory with the owner's name such as vlad-planets (replace vlad with owner's name).

kekoziar:
@spco, I agree with @pagrubel that this clarification is a good idea.  However, the sentence "The Collaborator doesn’t want to overwrite her own version of planets.git, so needs to clone the Owner’s repository to a different location to her own." is a little clunky.

Perhaps it can be revised to say 
"The Collaborator doesn’t want to overwrite her own version of planets.git, so will clone the Owner’s repository to a new folder called `vlad-planets` which will be created with the `git clone` command."

What do you think?

_episodes/08-collab.md
```diff
@@ -36,7 +36,9 @@ needs to go to [https://github.com/notifications](https://github.com/notificatio
 Once there she can accept access to the Owner's repo.
 
 Next, the Collaborator needs to download a copy of the Owner's repository to her
- machine. This is called "cloning a repo". To clone the Owner's repo into
+ machine. This is called "cloning a repo". The Collaborator doesn't want to overwrite
+ her own version of `planets.git`, so needs to clone the Owner's repository to a 
+ different location to her own. To clone the Owner's repo into
```
kekoziar, line 41: How about 

>  machine. This is called "cloning a repo". The Collaborator doesn't want to overwrite
 her own version of `planets.git`, so needs to clone the Owner's repository to a 
 different location than her own repository with the same name. To clone the Owner's repo into

?

kekoziar (commented):
Changes related to https://github.com/swcarpentry/git-novice/pull/740 (which will have caused a merge conflict).  It would be nice to make sure the whole section flows easily.

spco:
Merge conflict resolved - hopefully this reads more smoothly.

kekoziar (approved):
Changes and clarification read well. Conflict resolved.  Approved.

PR #574, opened 2018-11-15
Updated the index.md to remove unnecessary comma

zipho:
Removed an unnecessary comma with the documentation


nhejazi:
I agree with @kekoziar on this. There's no official grammatical style used in the Carpentries lessons and removing that comma does conflict with several accepted styles, as pointed out above. Thank you, @zipho, for the contribution but perhaps this may not get merged.

kekoziar:
Closing PR per above reasons.

PR #575, opened 2018-11-29
Episode 4: make Directories example more explicit

mtorpey:
Directories are an important concept in git, and this exercise is a useful one, but it can be improved.

This commit improves the example in two ways:
* Firstly, we now use the name "spaceships" for the directory instead of simply "directory"; this makes it clear to learners which part of the command is syntax and which is a name.
* Secondly, we add an extra code block that explicitly creates files and adds a directory.  The final `git status` command shows the user that the two new files have been added.

This was done as a contribution for Instructor Training Checkout.

### DIFF
--- a/_episodes/04-changes.md
+++ b/_episodes/04-changes.md
@@ -581,14 +581,14 @@ Date:   Thu Aug 22 09:51:46 2013 -0400
 >    Try it for yourself:
 >
 >    ~~~
->    $ mkdir directory
+>    $ mkdir spaceships
 >    $ git status
->    $ git add directory
+>    $ git add spaceships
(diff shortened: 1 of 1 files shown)


alex-konovalov:
Thanks - both points look like useful improvements to me. This will have to wait until one of the maintainers will look at it.

munkm:
Thanks @mtorpey for the PR and @alex-konovalov for the review! I really like this change. The spaceships folder (🚀 🚀 🚀) thematically goes with the rest of the lesson and it's less ambiguous than making a directory called `directory`. 

I previewed your changes and they render properly as well on my machine.

PR #581, opened 2018-12-14
Fixes #352 removing unnecessary update to mars.txt

### LINKED ISSUE #352: Exploring History actually involves losing a change
`05-history Exploring History` starts by adding another line to `mars.txt`:

> An ill-considered change

but erroneously only lists that new line in the output of the first diff. This change is never committed.

Later, `mars.txt` is overwritten and then recovered from the current HEAD. In effect, this deletes the "An ill-considered change" line from the file.

Solution could be to rewrite the first part of `05-history` to not add another change.

martinosorb:
Fixes #352 in the way suggested by @billsacks's points 1 and 3 (see Issue). I did not implement point 2, since I felt this is a different issue, which can be debated, but is not part of the bug. IMHO it would also be slightly confusing, as it takes attention away from the fact that we're working on that final line, but that's arguable.

### DIFF
--- a/_episodes/05-history.md
+++ b/_episodes/05-history.md
@@ -22,7 +22,7 @@ directory by using the identifier `HEAD`.
 
 We've been adding one line at a time to `mars.txt`, so it's easy to track our
 progress by looking, so let's do that using our `HEAD`s.  Before we start,
-let's make a change to `mars.txt`.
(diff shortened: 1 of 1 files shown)


kekoziar:
Thank you for implementing the suggestion. I always did a double-take at that section when delivering the workshop.

Issue #579, opened 2018-12-13
"config --edit" to both check editor config & configure more conveniently


katrinleinweber:
Would it make sense to update the section https://github.com/swcarpentry/git-novice/blame/0f775ae563a09163984e134b068d56928104ee29/_episodes/02-setup.md to something like:

> Let's test and review the settings we made just now. Type `git config --global --edit` to open your git configuration in the text editor that you just configured.

Related to https://github.com/swcarpentry/git-novice/issues/360, because it could prepare learners for switching between CLI & editor during git operations.

tkphd:
Possibly useful, as it teaches an alternative entry point into the config.
In a workshop I taught last week, I did something similar: to teach that the config is just a file, I had students `nano ~/.gitconfig` to see the options we had just set. This also applies to local `.git/config` files.

As a test to make sure the config was successful, so helpers can address misconfigurations, this would make sense. On the other hand, committing ourselves and our helpers to resolve configuration difficulties with the myriad possible editors might be a distracting scope creep for this lesson.

katrinleinweber:
Cool, thanks for sharing your teaching example :-) About "distracting": Would #671 help reduce that risk in your view?

However, since our [only](https://github.com/swcarpentry/git-novice/blame/04ffeff2715ba95433751f6bb0b2df978d6ed3af/_episodes/02-setup.md#L27-L28) [config](https://github.com/swcarpentry/git-novice/blame/04ffeff2715ba95433751f6bb0b2df978d6ed3af/_episodes/02-setup.md#L74-L76) instructions are crucial for the rest of the lessons, helpers will have to debug problems anyway, won't they? IMHO, this is a question of cleverly ordering instructions and checks so that errors do pop-up as quickly as possible.

PR #587, opened 2019-01-15
Test core.editor config immediately by listing it in there

katrinleinweber:
As suggested in #579.

### DIFF
--- a/_episodes/02-setup.md
+++ b/_episodes/02-setup.md
@@ -108,7 +108,8 @@ $ git config --global --edit
 ~~~
 {: .language-bash}
 
-However, it's safer to only view the configuration with:
+Let's close the file without making any additional changes.  Remember, since typos in the config file will cause 
+issues, it's safer to view the configuration with:
 
 ~~~
 $ git config --list
--- a/_episodes/02-setup.md
+++ b/_episodes/02-setup.md
@@ -101,15 +101,12 @@ your changes, press <kbd>Esc</kbd> then type `:q!` and hit <kbd>Return</kbd>.
 The four commands we just ran above only need to be run once: the flag `--global` tells Git
 to use the settings for every project, in your user account, on this computer.
 
-You can check your settings at any time:
+You can check your settings at any time, change them and add new ones more comfortably than through the command line:
 
 ~~~
-$ git config --list
+$ git config --global --edit
 ~~~
 {: .language-bash}
-
-You can change your configuration as many times as you want: use the
-same commands to choose another editor or update your email address.
 
 > ## Proxy
 >
--- a/_episodes/02-setup.md
+++ b/_episodes/02-setup.md
@@ -101,12 +101,23 @@ your changes, press <kbd>Esc</kbd> then type `:q!` and hit <kbd>Return</kbd>.
 The four commands we just ran above only need to be run once: the flag `--global` tells Git
 to use the settings for every project, in your user account, on this computer.
 
-You can check your settings at any time, change them and add new ones more comfortably than through the command line:
+Let's review those settings and test our `core.editor` right away:
 
 ~~~
 $ git config --global --edit
 ~~~
 {: .language-bash}
+
+However, it's safer to only view the configuration with:
+
+~~~
+$ git config --list
+~~~
+{: .language-bash}
+
+And if necessary, change your configuration using the
+same commands to choose another editor or update your email address.
+This can be done as many times as you want.
 
 > ## Proxy
 >


munkm:
Hi @katrinleinweber, thank you for both the issue and the PR to resolve it! I like that you've added an additional suggestion of how to pop into an editor to look at the config settings that are set. However, I'm a little hesitant to remove how to check the git config settings from the command line as is done in this PR. I do think it's reasonable to check config settings from the command line with `git config --list` or `git config -l`. Also, sometimes we just want to check our config settings but opening an editor to the file somewhat implies that we intend to make a change. I think that this would be a good supplement to listing from the command line, but I don't think we should delete the command-line list.

How would people feel if we modified this pr to have something like: 

```
You can check your settings at any time: 
~~~
$ git config --list	
~~~
{: .language-bash}

You can change your configuration as many times as you want: use the	
same commands to choose another editor or update your email address.

Alternatively, if you're more comfortable using an editor than reentering the command-line changes you can have git enter you into a core editor to modify the config file that stores your config settings. To do that, you'd do:

~~~
$ git config --global --edit
~~~	
{: .language-bash}
```

katrinleinweber:
The [CG](https://github.com/swcarpentry/git-novice/blob/gh-pages/CONTRIBUTING.md#what-not-to-contribute) states "material", not "whole new sections. My `[new concept]` addendum was just an example.

> if the core editor isn't set properly then entering into the git config could be problematic.

Exactly that's why it should be tested (nost just viewed) immediately, not remain untested until some advanced operation launches learners into vi(m) or an error message unexpectedly ;-) This is my core argument in this PR.

Assuming they use `--list` after making a copy-paste or typing error when setting the config, isn't it less likely to be spotted by viewing only.

> does't show learners the workflow where you'd want to check config settings in a way that's completely separate from editing a file [...] not default to modifying the file that has them.

Opening a file primarily means _viewing_ its content. That is separate from editing it. Which in turn is separate from saving edits. Which can be undone. So, we're at least 2 steps removed from the risk of learners breaking their git config, while ensuring the discovery & repair of the problems right then & there.

kekoziar:
I agree that `git config --list` should remain in the lesson.  I have seen config issues where `git config --global --edit` wouldn't open the file, but examining the settings using `git config --list` helped identify the problem.  I think it's important for new learners to know the basic commands.

I also like the idea of *testing* the `core.editor` setting - mostly because I've also seen windows config settings fail during a workshop, and it would have been easier to fix at that time than later in the lesson.  The modification @munkm proposed allows testing the `core editor`, but it needs an addendum or to be reworked.  An issue is `git config --global --edit` shows different output than `git config --list`.  This is because `git config --global --edit` only shows the configurations added using the `--global` tag, and not other configuration settings which are either repository or system based.  This is where we get into scope creep.

While I don't think adding a simple "let's test the `core.editor` setting" line is adding a new concept, the explanation behind "let's check and change configuration settings using the `core.editor`" can start to creep into a new topic. 

That said, I think `git config --list` needs to remain in the lesson because it is a basic command.  If `git config --global --edit` is added, it should be immediately after the `core.editor` is added and only to test that the `core.editor` configuration works. Anything beyond that I think will end up being a more advanced topic and out of scope of this lesson.

katrinleinweber:
Thanks for the feedback and sorry that I didn't find time to incorporate it before. How about this version?

_episodes/02-setup.md
```diff
@@ -101,15 +101,23 @@ your changes, press <kbd>Esc</kbd> then type `:q!` and hit <kbd>Return</kbd>.
...
 
-You can check your settings at any time:
+Let's review those settings and test our `core.editor` right away:
+
+~~~
+$ git config --global --edit
+~~~
+{: .language-bash}
+
+However, it's safer to only view the configuration with:
```
kekoziar, line 111: How about
> However, since typos in the config file will cause issues, it's safer to only view the configuration with:

?

kekoziar (changes requested):
Branch needs to be update prior to merging.

kekoziar (commented):
Also, related to https://github.com/swcarpentry/git-novice/issues/579

_episodes/02-setup.md
katrinleinweber: Hi @kekoziar 👋 You commented on an older version of the diff. Is the current one also in need of changes? If so, please note that I've activated `Allow edits by maintainers`. Please feel free to apply your favourite wording directly. Cheers :-)

katrinleinweber:
> …katrinleinweber needs to update her branch and resubmit for this pull request to proceed?

If the GitHub checks can't be triggered otherwise, I'm fine with doing so. Since there are no merge conflicts, it wouldn't strictly be necessary.

kekoziar (approved):
looks good and renders fine.  Thank you everyone for this update!

PR #588, opened 2019-01-15
Complete `git rm` command

katrinleinweber:
This was proposed in #558, but not included in #560. Thus, the example was out-of-sync with the explanation above it, isn't it?

@krishnanj & @nhejazi: Please also note my [diff comments in #560](https://github.com/swcarpentry/git-novice/pull/560/files). I enabled the `Allow edits from maintainers` option in case more commits should be pushed into this PR.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -146,7 +146,7 @@ wording of the output might be slightly different.
 > > ### Background
 > > Removing files from a git repository needs to be done with caution. To remove files from the working tree and not from your working directory, use
 > > ~~~
-> > $ rm filename
+> > $ git rm filename
 > > ~~~
 > > {: .language-bash}
 > > 


kekoziar:
Thank you for the contribution. I think there is a misunderstanding here.  In issue #558, @nhejazi wrote (and I agree), "this is a fairly introductory lesson and I think `git rm` and related commands might be out of scope for this level of introduction. It would be nice to see a lesson covering some of these concepts but it might fall in an appendix section of the lesson or under advanced topics."   

So, the original proposal in #558 to use git command `git rm` was not approved, but the concept of further explanation of a remove command was approved and #560 changed the syntax to use the bash command `rm`.

katrinleinweber:
Thanks for trying to clarify, but the lesson wording is still very misunderstandable: `rm filename` _does_ remove the file from the WD.

I admit to being unsure where the difference between `working tree` and `working directory` should be, but one or more of these 4 terms/commands (above, plus `git repository`) is definitely out of place in that sentence, and contradicts [shell-novice/03](https://swcarpentry.github.io/shell-novice/03-create/index.html#removing-files-and-directories).

PR #591, opened 2019-01-27
Improve accuracy of language

caseyyoungflesh:
Originally statement:
"Create a new Git repository on your computer called bio."

Technically, I don't think the Git repository has a name on the local machine (though maybe others disagree) - rather there is a repository in a directory called 'bio' (however, on Github I would definitely say the repo has a name). Changed language in the exercise to reflect this.


munkm:
Hi @caseyyoungflesh! Thanks so much for your PR and your thoughts about the wording here. I'd like other maintainers to weigh in since this is a bit subtle, but I'm not entirely sure that I agree with this change. I agree that in a local directory structure the repository lives inside the `bio` folder. However, if we're dealing with any remote we'd say that we're cloning/dealing with the bio repository (e.g on the remote server `myserver` we'd clone the bio repository with `myserver/bio.git`).  The difference in this language could be confusing to new learners, and distinguishing why what a repository is called locally vs. remotely is different is probably too advanced for the lesson. 

That said, I like that you distinguish the creation of a new directory/folder and the initialization of the repository on line 736 as two separate actions. I would be willing to merge that, but not the modified title of the exercise.

caseyyoungflesh:
@munkm makes sense! I could definitely see the argument for changing the text but not the title.

kekoziar:
Hi, @caseyyoungflesh, thanks for the PR.  I agree with what @munkm wrote regarding this proposed change, especially the need to avoid confusing new learners.  When I refer to a repository on my local machine, it is by the folder name; to me, they are one and the same regardless of if I have connected my local repo with a remote.

Other questions we might ask are, are the learners at a sufficient place to understand that when someone asks to create a `bio` repository, it is a two step process (folder creation then repo init); and is the knowledge that it is a two step process something we as instructors want the exercise to assess?  As the exercise is now written, it assesses whether the learners understand that creating a repo is a two step process (or that the repo needs to be in a folder), while the rewrite lists the individual steps and only assesses if the learners recall the specific commands.  I prefer assessing if learners can implement the process rather than memory recall of commands.

caseyyoungflesh:
@kekoziar I see where you're coming from here. I find the existing language more confusing than helpful (likely because of my mental map of all of this), but fully understand that I might be in the minority!

kekoziar:
The wording we use is actually very similar to the book at [Chapter 2.1](https://git-scm.com/book/en/v2/Git-Basics-Getting-a-Git-Repository).
```
You typically obtain a Git repository in one of two ways:

You can take a local directory that is currently not under version control, and turn it into a Git repository, or [...]
```
turn a local directory into a repository; make a folder a repository; refer to a repository by the folder/directory name.  It's easier for learners to understand this new concept by first couching it in language that is familiar, then explaining the technicalities.

Issue #590, opened 2019-01-26
Fix git install instruction link in lesson 14


munkm:
As mentioned in swcarpentry/git-novice#531 the URL to download and install git in Lesson 14 should be consistent with the rest of git-novice. Right now this lesson says 

`If there is no version of Git on your computer, please either install [Git](https://git-scm.com/downloads/)`

This should be changed to:

`If there is no version of Git on your computer, please either install [Git](https://swcarpentry.github.io/git-novice/setup.html)`

PR #593, opened 2019-01-28
changed Git install URL to be consistent with other episodes

TraciPopejoy:
I changed the URL fir Git install so it is consistent among episodes, as suggested here: https://github.com/swcarpentry/git-novice/issues/590#issue-403400970 Does the other URL need to be edited as well?

The website it leads to is a little strange. Is the learner to change directory to desktop and then install or install and then change directory? It might be evident in the normal flow of a lesson.

### DIFF
--- a/_episodes/14-supplemental-rstudio.md
+++ b/_episodes/14-supplemental-rstudio.md
@@ -47,8 +47,9 @@ our computer, we choose the option "Existing Directory":
 > - `where git` (Windows)
 >
 > If there is no version of Git on your computer, 
-please either install 
-[Git](https://swcarpentry.github.io/git-novice/setup.html)
+please follow the 
+[Git installation 
+instructions](https://swcarpentry.github.io/git-novice/setup.html)
 > in the setup of this lesson to install Git now. Next open your shell or command prompt 
 > and type `which git` (Mac, Linux), or `where git` (Windows).
 > Copy the path to the git executable.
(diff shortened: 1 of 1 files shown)


munkm:
Hi @TraciPopejoy! Thank you so much for the Pull Request! 

You've brought up a really good point here. The page that I linked is to the setup instructions for the lesson, which links to the git install directions in the workshop template and then says to move into one's `Desktop` folder. At this point in the lesson there's no need to move into the `Desktop` folder. However, executing the commands in the subsequent sentences (like `which git` or `where git`) should work in any directory on one's machine. Maybe it would be clearer to say that the page we're linking is the setup for the lesson. 
The second URL about github desktop points to something that we don't really support in this lesson either, so it should probably be removed. Would you be willing to update the sentence from 

```
If there is no version of Git on your computer, please either install 
[Git](https://swcarpentry.github.io/git-novice/setup.html) or [GitHub](https://desktop.github.com/) now. Next open your shell or command prompt and type `which git` (Mac, Linux), or `where git` (Windows).
```

to something like 

```If there is no version of Git on your computer, please follow the [Git installation instructions](https://swcarpentry.github.io/git-novice/setup.html) in the setup of this lesson to install 
 Git now. Next open your shell or command prompt and type `which git` (Mac, Linux), or `where git` (Windows). ```

If you think it's clearer to link directly to the git install directions that are linked on that page, then feel free to use that link instead and keep the original wording.

TraciPopejoy:
I think the install page is fine. Thank you for suggesting these changes though. I'm sorry I have responded so slowly, but I hope this helps.

munkm:
Awesome! Thank you so much @TraciPopejoy! Your PR resolves swcarpentry/git-novice#590.

PR #595, opened 2019-02-07
Propose a efficient way to commit more files

nanhung:


kekoziar:
I agree with @tkphd's comments.  This command could add confusion by introducing a more advanced command and concepts at a time when the learners are working through introductory concepts of the git process. Thank you for the contribution, but this PR will be closed because it doesn't fit with the pedagogy of the lesson.

nanhung:
Gotcha! Thanks for y'all explanations :).

PR #598, opened 2019-02-23
commit staged spaceship files

### LINKED ISSUE #597: spaceships in Episode are added but never committed.
In Episode 4, we do the following:
```
$ touch spaceships/apollo-11 spaceships/sputnik-1
$ git status
$ git add spaceships
$ git status
```
but we never commit this.  This leaves these files in the staging area.  Later as we add more to the repo, we never discuss these uncommitted files.  Perhaps they should just be committed after adding here?

zingale:
In lesson 4, we add spaceship files but we never commit them.  This leaves them in the staging area, which confuses the subsequent instructions.

closes #597

### DIFF
--- a/_episodes/04-changes.md
+++ b/_episodes/04-changes.md
@@ -613,6 +613,13 @@ Date:   Thu Aug 22 09:51:46 2013 -0400
 >    ~~~
 >    {: .language-bash}
 >
+>    Before moving on, we will commit these changes.
+>
+>    ---
+>    $ git commit -m "Add some initial thoughts on spaceships"
+>    ---
+>    {: .language-bash}
+>
 {: .callout}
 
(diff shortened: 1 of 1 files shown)


_episodes/04-changes.md
```diff
@@ -613,6 +613,13 @@ Date:   Thu Aug 22 09:51:46 2013 -0400
 >    ~~~
 >    {: .language-bash}
 >
+>    Before moving on, we will commit these changes.
+>
+>    ---
```
munkm: I think this (and line 620) needs to be changed from `---` to `~~~` for the git commit command to render properly.

zingale: thanks.  it should be fixed now.

munkm: That was so fast! Thank you!

PR #601, opened 2019-03-15
Update 01-basics.md

aarontaycheehsien:
slight rewording from "but later ruin it" to "but you ruined it later"


kekoziar:
@aarontaycheehsien, thanks for the suggestion.  I don't exactly agree with it.  There wasn't ambiguity in the original sentence, and the change adds needless words.  While The Carpentries doesn't have an official adopted style, I defer Strunk & White for composition style.

PR #602, opened 2019-03-15
Add link for sync from Github to Figshare

aarontaycheehsien:
added alternative to Zenodo

---


kekoziar:
[Figshare only allows CC-0 with free accounts](https://knowledge.figshare.com/articles/item/copyright-and-licence-policy), while Zenodo supports all licenses. If a link to Figshare is added, I think this should be explicit, since it is not explicit on their website, but difficult to find. Unless you can find examples and terms on Figshare that explicitly states free accounts can have other licenses.

kekoziar:
I'm going to close this because changes requested by maintainers were never implemented, and PR appears abandoned. Please feel free to reopen or submit a new PR with requested changes (latter preferable as it resolves conflicts, and allow author to take into account any changes to the lesson text since this PR).

Issue #508, opened 2018-05-06
Opening of the Version Control lesson starts with an assumption


iglpdc:
(From https://github.com/swcarpentry/shell-novice/issues/774)

The first line of the Version Control with GIT Tracking Changes starts at a point that stopped me.
It says:
"First let’s make sure we’re still in the right directory. You should be in the planets directory.
$ pwd"

I did not know where to find the "planets" directory. I would like to "lead in" statement that says: "Go to the directory where you will find the planets directory"
and add a link.

(cc @sandycampbell)

iglpdc:
Thanks, @sandycampbell !

I agree that this first sentence can be confussing. What do you think about keeping the text as it is, but changing the command?

> First let’s make sure we’re still in the right directory. You should be in the `planets` directory.
```
$ cd 
$ cd /Desktop/planets
```

This will work in all the operating systems and in the previous episode the `planets` directory is created in the `Desktop`, so everybody should have it there.

Would you like to submit a PR with this change?

rsuchecki:
There are a few issues with this issue:

1. The first line in [4. Tracking Changes](https://github.com/swcarpentry/git-novice/blob/879c57bac9a9ba344f3154aeba8f5a16e1f517bd/_episodes/04-changes.md#L10-15) follows directly from the work in [3. Creating a Repository](https://github.com/swcarpentry/git-novice/blob/879c57bac9a9ba344f3154aeba8f5a16e1f517bd/_episodes/03-create.md) so unless someone skips the previous section it is all pretty clear.
2. The proposed solution appears to be incorrect, as it expects the `Desktop` directly under the root `/` directory (`/Desktop/planets`) which is inconsistent with the rest of the lesson. The intended directory location is `~/Desktop/planets`, that is,  `/home/${USER}/Desktop/planets` 

**Conclusion**: this issue should either be 
* closed with a wont-fix label 
or 
* the first line could be kept and the following 6 lines  replaced with 

```
cd ~/Desktop/planets
```
Lines to be removed in this case:

```
$ pwd
/home/vlad/Desktop/planets
```

> If you are still in moons, navigate back up to planets

```
$ pwd
/home/vlad/Desktop/planets/moons
$ cd ..
```

rainsworth:
I think this issue can be closed because it was resolved in #605 ?

munkm:
Yes! Thank you!

PR #605, opened 2019-03-28
_episodes/04-changes.md: correct starting directory verbage

prwolfe:
This is to resolve issue #508

Please delete the text below before submitting your contribution. 

This makes sense to me base on both my experience and the discussion - let me know what you think.


munkm:
I'm going to see what other maintainers think of this, but I like this variation you've proposed. It's a cleaner way to ensure all users get back to the planets directory.

kekoziar:
@munkm I agree with you. I like this change!

prwolfe:
Thanks!

PR #616, opened 2019-04-30
Adjusting the order of commands in "Tracking Changes" (lesson 4)

NZR:
Hello everyone, 

Please find attached 2 commits to adjust the order of the steps in which commands are run to check the existence of the "mars.txt" file. 

The lesson suggest creating the file, check its content, and check it exists (using ls) - according to the sequence of command - not the text! 

I re-order the step as follows: 
- create the file
- check it exists (using ls)
- check its content (using cat)

All commands were there before, I just re-ordered them and added a little bit of text to make the transition between the command more fluid.


kekoziar:
Hi, @NZR.  Thanks for the suggestion.  Can you change line 72 of your PR from `mars.txt moons/` to `mars.txt`?  Generally speaking, output of commands that are executed during the lesson does not include items that were added from exercises.  Not all instructors teach all exercises (nor are they expected to), and the `moons` directory was created in an exercise.

NZR:
Hello ! 

Thanks for taking the time to review the submission. 

I am on the fence regarding the "moons/" folder. I think it should stay because it is mentioned earlier in the episode. 

In the first steps of the episode, we first check where we are in the file tree. One of the possibility (exposed in the episode) is that the learner is still in the "moons/" subfolder, and the instructions are given on how to get back to the root project folder ("planets"). 

Therefore, there should be no scenarios in this episode where the "moons/" folder does not exist (either because of previous episode, or because of what is stated in the early phases on this episode).

But that's my only argument in favor on keeping the "moons/" folder in the output of the example. So I am flexible ;) 

Let me know and if you are not convince I will change it back. 

Thanks!

kekoziar:
The section you mention is the only place outside of the exercise where the `moons/` folder is mention. I actually don't agree that it should be there, and believe that section can be reworded to avoid it.  When I teach this section, a `moons/` folder is rarely created because we use that particular exercise as a thought experiment. 

Regardless, **lesson** command output does not include items added from exercises. Not all instructors teach all exercises (nor are they expected to), and the `moons/` directory was created in an exercise.  We maintain this guideline because adding output from exercises will make the lesson unwieldy, and difficult for instructors to use when teaching the lesson.  It is far easier for instructors to mentally add in output from exercises they executed in class, than see cluttered output in the lesson, and try to differentiate lesson output from exercises they did not use.

NZR:
Hello!

Thanks for your reply. As I mentioned, I don't have any strong feeling about this point. I've just updated the pull request (commit f7aa767), removing the references to the moon folder in the command output. 

This should be alright now! 

Thanks again, and have a great day! 
- Nico

Issue #652, opened 2019-05-27
All-in-one-page broken


iamc:
Episodes -> "all in one page" broken.

fmichonneau:
to fix it, please replace `aio.md` in this repo with this file: https://github.com/carpentries/styles/blob/gh-pages/bin/boilerplate/aio.md

kekoziar:
Thank you for your submission. Since an updated version of the lesson style has been implemented, fixing the original problem this PR addressed, this PR will now be closed.

PR #653, opened 2019-05-28
Update aoi.md from boilerplate.

iamc:
Fixes https://github.com/swcarpentry/git-novice/issues/652

### DIFF
--- a/aio.md
+++ b/aio.md
@@ -1,6 +1,4 @@
 ---
-layout: page 
-permalink: /aio/
 ---
 <script>
   window.onload = function() {


fmichonneau:
Thank you for contribution @iamc!

It looks like you correctly updated the yaml header to match the newer version of the file that is in the styles repository, however the content of the body of the file has also changed. Could you please update your pull request to also include the updated code in the file? thanks!

iamc:
I updated from the file `bin/boilerplate/aio.m` in this same repo and all-in-one generation seems to work ok.

I did try replacing, as by @fmichonneau suggestion, with the one in https://github.com/carpentries/styles/blob/gh-pages/bin/boilerplate/aio.md but this does not generate the all-in-one page, but tries to download de aio.md file instead. At least when serving the page locally in my computer.

kekoziar:
Thank you for your submission. Since an updated version of the lesson style has been implemented, fixing the original problem this PR addressed, this PR will now be closed.

Issue #625, opened 2019-05-06
Episode 7 - Remotes in GitHub: Update object/bytes references


kekoziar:
Regarding
~~~
Counting objects: 9, done.
Delta compression using up to 4 threads.
Compressing objects: 100% (6/6), done.
Writing objects: 100% (9/9), 821 bytes, done.
Total 9 (delta 2), reused 0 (delta 0)
To https://github.com/vlad/planets
 * [new branch]      master -> master
Branch master set up to track remote branch master from origin.
~~~
Because of addition to the lesson (specifically, adding spaceships) objects have changed such that the counts are now

~~~
Enumerating objects: 16, done.
Counting objects: 100% (16/16), done.
Delta compression using up to 8 threads
Compressing objects: 100% (11/11), done.
Writing objects: 100% (16/16), 1.45 KiB | 372.00 KiB/s, done.
Total 16 (delta 2), reused 0 (delta 0)
remote: Resolving deltas: 100% (2/2), done.
To https://github.com/vlad/planets.git
 * [new branch]      master -> master
 ~~~

While the output will be different based on what exercises have been completed, IMO we should have the main text of the lesson be as true as possible to the output of the lesson sans exercises.

Also, my output does not have the last line, 
 "Branch master set up to track remote branch master from origin"
 can someone verify if this is a change to git or something else?

munkm:
I did a little dummy example and found that I got a variant of the last line if I executed 
```
$ git push -u origin master
Counting objects: 6, done.
Delta compression using up to 4 threads.
Compressing objects: 100% (3/3), done.
Writing objects: 100% (6/6), 420 bytes | 420.00 KiB/s, done.
Total 6 (delta 0), reused 0 (delta 0)
To github.com:munkm/testrepo.git
 * [new branch]      master -> master
Branch master set up to track remote branch master from origin by rebasing.
```

Whereas when I didn't use the upstream option I didn't get that message

```
$ git push origin master
Counting objects: 6, done.
Delta compression using up to 4 threads.
Compressing objects: 100% (3/3), done.
Writing objects: 100% (6/6), 420 bytes | 420.00 KiB/s, done.
Total 6 (delta 0), reused 0 (delta 0)
To github.com:munkm/testrepo.git
 * [new branch]      master -> master 
```

kekoziar:
Hmmm.  Upstream isn't introduced until later in that episode in the call-out, **The ‘-u’ Flag**.

munkm:
[...]

PR #655, opened 2019-05-29
update episode 7 to reflect current output per issue #625.

kekoziar:
Update episode 7 to reflect current output per issue #625. Also, update instructor guide, as suggested to call out differences in output if using github suggested syntax and push includes upstream (e.g. `git push -u origin master`).

### DIFF
--- a/_extras/guide.md
+++ b/_extras/guide.md
@@ -173,6 +173,10 @@ particular set of files in `.gitignore`.
 *   It is very useful to draw a diagram showing the different repositories
     involved.
 
+*   Learners using syntax from GitHub, `git push -u origin master`, will have slightly different 
+    output, including the line `Branch master set up to track remote branch master from origin by 
+    rebasing.`
+
 ## [Collaborating]({{ page.root }}/08-collab/)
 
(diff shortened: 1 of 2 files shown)


_extras/guide.md
```diff
@@ -173,6 +173,10 @@ particular set of files in `.gitignore`.
 *   It is very useful to draw a diagram showing the different repositories
     involved.
 
+*   Learners using syntax from GitHub, `git push -u origin master`, will have slightly different
```
munkm: I like this addition to the guide, but it might be helpful if there were slightly more context. Maybe something like:

> When pushing to a remote, the output from Git can vary slightly depending on what leaners execute. The lesson displays the output from git if a learner executes `git push origin master`. However, some learners might use syntax suggested by GitHub for pushing to a remote with an existing repository, which is `git push -u origin master` ... 

and then include everything you've put! What do you think?

kekoziar: Sounds good!

kekoziar: I updated the instructor guide to add your suggestion.

munkm:
Fantastic! This looks great.

PR #660, opened 2019-05-30
added registry of Data Repositories

eturnator:
This commit is to fulfill part of teacher training requirement for the carpentries.

### DIFF
--- a/_episodes/07-github.md
+++ b/_episodes/07-github.md
@@ -149,7 +149,7 @@ Branch master set up to track remote branch master from origin.
 >
 > ~~~
 > $ git config --global http.proxy http://user:password@proxy.url
-> $ git config --global https.proxy http://user:password@proxy.url
+> $ git config --global https.proxy https://user:password@proxy.url
 > ~~~
 > {: .language-bash}
 >
--- a/_episodes/10-open.md
+++ b/_episodes/10-open.md
@@ -112,7 +112,7 @@ by acting as a shareable electronic lab notebook for computational work:
 >
 > Surf the internet for a couple of minutes and check out the data repositories
 > mentioned above: [Figshare](https://figshare.com/), [Zenodo](https://zenodo.org),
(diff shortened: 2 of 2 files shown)


kekoziar:
Thanks for the suggested change, @eturnator. I'm not convinced that adding re3data.org here is the best way to introduce the concept of data repositories to new learners.  The Nature article is very good because it contains a curated list of repositories that _"have been evaluated to ensure that they meet [Nature's] requirements for data access, preservation and stability."_  FAIRsharing.org and re3data.org are listed on the Nature page, but toward the bottom and after some explanation and listing of other repositories.  The page is also good because it describes the type of data submitted to the disciplinary repositories.  

In contrast, re3data.org contains links to over 2,000 data repositories with 27 different filters used to narrow selections. IMO, it is difficult for a learner new to the concept of data repositories and open data to narrow and select a trustworthy and persistent repository; that takes intermediate/advanced skills and knowledge.  I think it’s better to show examples of good data repositories, than link to a large index that could either overwhelm new learners and/or cause new learners to overlook advantages and drawbacks to different data repositories.

Is there a use case or situation that motivated this PR?

eturnator:
I agree that the Nature article is a good introduction (with the important caveat that despite the good quality of the content of that article Nature itself is not an open and freely available resource) . Note however, that the examples from Nature are all from hard science fields; not everyone coming to these pages will be from those science fields. So the re3.org site complements that information and makes other repositories accessible and this might actually benefit new learners rather than overwhelm them.  It's filter makes it very easy for new learners to identify repositories that they might need, be it field specific, region specific or otherwise. We might need to change the wording or place it after the Nature articles but I think it new learners should be made aware of its existence. 

All my best,
Ece

kekoziar:
Hi, @eturnator.  Will you please submit separate PRs for different issues instead of adding multiple issues or episode changes to one PR?  Thanks.

PR #661, opened 2019-06-02
GitHub’s urls to resources have changes

wna-se:


kekoziar:
@wna-se, thank you for finding and fixing this.  Since the [parent page](https://github.github.com/training-kit/) for the cheat sheet contains cheat sheets in many languages, it might be worth it to change the lesson text slightly and refer to the parent page.  Will you make those changes?

kekoziar:
I verified the submitted change works.  Since this fixes a now broken link, I will merge the PR as submitted, and create a new issue with the suggested change.

wna-se:
@kekoziar, I'm pretty swamped this week but If the new issue is still open next week there's a good chance I could pick it up and make the changes

PR #662, opened 2019-06-04
Upgrade style 90f06fc

fmichonneau:
here is a PR to upgrade to the latest version of the lesson template.


kekoziar:
Hi, @fmichonneau.  I have a question about this style upgrade.  When I render this PR of the git-novice lesson locally, the font size in callout boxes, exercises/challenges, and overview is markedly larger than the lesson text.  I have replicated this on two different machines. Below is a screenshot of the git-novice lesson compared to the r-novice-inflammation lesson (which is upgraded to style 9.5.3).  I also cloned r-novice-inflammation and rendered it locally, and it does not show a font size difference.  Is this font size difference in the PR correct?
compareStyles_git-novice_v_r-inflammation-annotated

Also, I noticed the code boxes do not always have a complete outline.  I tested this with local git-novice and the live r-novice-inflammation, and both exhibit the behavior. It seems to be a browser issue, where the code boxes render with a complete outline on FireFox, but not on Chrome nor Edge.  I am less worried about this, but thought you should know of the issue.

Can you help with the font size issue?  If you do not see this issue locally, will you share the version of ruby, jekyll, and bundler you are running?

fmichonneau:
@kekoziar the font issue was a bug that was just fixed in the styles repo (https://github.com/carpentries/styles/pull/414). I'm going to update this PR to include this fix and another issue regarding the contrast of the font for the code blocks. I opened an issue regarding the incomplete box borders so we can investigate further (carpentries/styles#415)

maxim-belkin:
@kekoziar @fmichonneau I can reproduce the problem with the border in Google Chrome on macOS.
Apparently, this is a known issue: https://stackoverflow.com/q/33737100

Perhaps, we should switch to 1px borders.

maxim-belkin:
Forgot to mention: currently, our borders in code blocks are 0.5px (and 5px on the left side).

kekoziar:
verified font issue fix. merging now.

PR #665, opened 2019-06-04
fix links

fmichonneau:
use the `{% link ... %}` notation so the paths are correct for offline use

### DIFF
--- a/_episodes/01-basics.md
+++ b/_episodes/01-basics.md
@@ -47,12 +47,13 @@ Unless multiple users make changes to the same section of the document - a confl
 ![Multiple Versions Can be Merged](../fig/merge.svg)
 
 A version control system is a tool that keeps track of these changes for us,
-effectively creating different versions of our files. It allows us to
-decide which changes will be made to the next version (each record of these changes is called a
-[commit]({{ page.root }}/reference#commit)), and keeps useful metadata about them. The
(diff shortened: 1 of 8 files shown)


kekoziar:
Hi, @fmichonneau.  Can you verify the change in episode 13 is supposed to be {{ swc_github }} and not {{ site.swc_github }}? My local rendering does not have a link for {{ swc_github }}

kekoziar:
An additional note that I checked all other changes, and fixed a couple of small errors.  The {{ swc_github }} question in the above comment is the only item that needs verification prior to merging.

munkm:
That item was also addressed in this PR: https://github.com/swcarpentry/git-novice/pull/649 and was a broken link in our lesson before that PR was merged.

kekoziar:
@munkm, I'd guess that the reversion to the syntax previous to PR #649 is a mistake, and the correct syntax is  {{ site.swc_github }}.  Thank you for clarifying!

kekoziar:
Verified link syntax fix.

kekoziar:
@fmichonneau, I checked this locally, but not live. The links are broken live =(

kekoziar:
@fmichonneau, nvm. The links are fixed. they were missing `{{ page.root }}` in front of the `{% link .. %}`

PR #666, opened 2019-06-05
Remove references to forking from Episode 8

### LINKED ISSUE #584: Collaborating episode references forking without explaining it
_episodes/08-collab.md references forking in the solution to the first challenge (`On GitHub, the Collaborator can go to their own fork of the repository` etc.) without ever previously mentioning forking, unless I am missing something (if I just missed this, please let me know!)

I would be glad to submit a PR with content to rectify this, but would first want to hear others' thoughts on whether forking should or should not be in the scope of this lesson.

kelly-sovacool:
Resolves #584

### DIFF
--- a/_episodes/08-collab.md
+++ b/_episodes/08-collab.md
@@ -194,14 +194,8 @@ GitHub) are back in sync.
 > > them. Then by running ```git diff master origin/master``` the Collaborator
 > > will see the changes output in the terminal.
 > >
-> > On GitHub, the Collaborator can go to their own fork of the repository and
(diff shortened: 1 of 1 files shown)


munkm:
Hi @kelly-sovacool! Thank you so much for fixing this and for your patience on getting a review. I like the changes you've made. Because both learners are using the same remote and haven't learned to branch, there isn't a way for them to compare changes using github. However, they can see what has been changed by inspecting the changes that have been pushed to github with the way you've proposed.

PR #667, opened 2019-06-20
Fixing broken links and the All-In-One page

jakelever:
Hi, this is just a collection of tiny fixes to get some broken links working. There were some "{{ page.root }}"s missing on various pages so the relative paths were broken for several things.

Also the All-In-One page is currently broken, so this includes a small change to fix that (by removing header information).


kekoziar:
Thank you for fixing the links. They worked locally, but not live.

PR #668, opened 2019-06-24
Adding an objective covered in the lesson

catsmith:
git status seems to me an important and useful enough command to be included in the lesson objectives.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -6,6 +6,7 @@ questions:
 - "Where does Git store information?"
 objectives:
 - "Create a local Git repository."
+- "Check the current status of a local Git repository."
 keypoints:
 - "`git init` initializes a repository."
 - "Git stores all of its repository data in the `.git` directory."


munkm:
@catsmith I think this is a great addition to the objectives of this lesson. It's also helpful for future lessons, so really emphasizing it in the lesson where it's introduced is a good idea. However, lesson 04 heavily emphasizes git status too, and the questions and keypoints in that lesson are "How do I check the status of my version control repository?" and "`git status` shows the status of a repository." , respectively. Maybe you could modify your keypoint here to be "Introduce how to check the status of a local Git repository."? What do you think? 

I think it might also be nice to update the question relating to git status in lesson 04 to reflect this too. Lesson 04 really shows how powerful git status is in checking our repository (and how it changes as we add and commit files), but since it is introduced in lesson 03 we may want to reword it slightly. I'd update it to something like "How does `git status` change as I make changes in my version control repository?".

munkm:
Hi @catsmith would you like any help updating these objectives slightly? I'm happy to help if you need it.

kekoziar (changes requested):
my main question is how does this pair with the in-depth use of status in episode 4?

emcaulay:
In reviewing the episode 3 today, I don't see any use of "git status" in this episode. I think this pull request has become outdated.

kekoziar:
this objective was subsequently added to Episode 4, which in depth uses `git status.` 
Closing.

PR #672, opened 2019-06-28
Update history episode with better terminology

jonandernovella:
This PR aims at improving the readability of the fifth chapter of the git-novice lessons. Specifically, I suggested two changes:

- Replacing `staged area` with `staging area`

and 

- Replacing `get rid of` with `discard`

I believe that this is more common terminology in the Git lingo. Let me know what you think.

### DIFF
--- a/_episodes/05-history.md
+++ b/_episodes/05-history.md
@@ -256,7 +256,7 @@ Changes to be committed:
 ~~~
 {: .output}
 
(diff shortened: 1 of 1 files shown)


drjarno:
This is a good idea. Especially the staging area bit. It's good to follow the terminology like you say.

PR #680, opened 2019-07-25
link to more detailed explanation of core.autocrlf configuration option

markmatney:
This PR replaces a link to supplementary documentation about the `core.autocrlf` config option that I found to be inadequate for a couple reasons:

- It doesn't offer much more than the inline explanation in the Carpentries lesson itself, in terms of justifying the config setting, beyond "for properly handling line endings". The new link concretely explains potential sources of the issue and describes the actual file transformations that take place with each setting.
- This issue is fundamentally about Git and not GitHub, so a link to the Git documentation is more appropriate.

### DIFF
--- a/_episodes/02-setup.md
+++ b/_episodes/02-setup.md
@@ -54,7 +54,7 @@ For these lessons, we will be interacting with [GitHub](https://github.com/) and
 > Because Git uses these characters to compare files,
 > it may cause unexpected issues when editing a file on different machines. 
 > Though it is beyond the scope of this lesson, you can read more about this issue 
-> [on this GitHub page](https://help.github.com/articles/dealing-with-line-endings/).
+> [in the Pro Git book](https://www.git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_core_autocrlf).
 {: .callout}
 >
 > You can change the way Git recognizes and encodes line endings
--- a/_extras/guide.md
+++ b/_extras/guide.md
@@ -58,6 +58,9 @@ working in teams or not, because it is
     these issues, since learners will almost certainly trip over them
     again.  If learners are running into line ending problems, GitHub
     has a [page][github-line-endings] that helps with troubleshooting.
+    Specifically, the [section on refreshing a repository][github-line-endings-refresh]
+    may be helpful if learners need to change the `core.autocrlf` setting
+    after already having made one or more commits.
 
 *   We don't use a Git GUI in these notes because we haven't found one that
     installs easily and runs reliably on the three major operating systems, and
@@ -302,7 +305,8 @@ web-hosted private repositories.
 [git-parable]: http://tom.preston-werner.com/2009/05/19/the-git-parable.html
 [github]: https://github.com/
 [github-gui]: https://git-scm.com/downloads/guis
-[github-line-endings]: https://help.github.com/articles/dealing-with-line-endings/#platform-all
+[github-line-endings]: https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings
+[github-line-endings-refresh]: https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings#refreshing-a-repository-after-changing-line-endings
 [github-privacy]: https://help.github.com/articles/keeping-your-email-address-private/
 [repos-in-repos]: https://github.com/swcarpentry/git-novice/issues/272
 [try-git]: https://try.github.io


drjarno:
I'm not so sure about this. While I agree that it is fundamentally a Git thing, the original link to the GitHub page gives a much better explanation compared to the rather terse version in the Git book even if there is some overlap with the lecture.

In addition, the GitHub link mentions practical tips such as how to refresh the repository without having to commit files that you didn't modify due to the new interpretation of the CRLF. This is not mentioned in the lecture, I think.

markmatney:
By "better", I mean better *in the context of the lesson* (as a linked reference that is supposed to provide "more information"), which I still think it is.

First, regarding

> the GitHub link mentions practical tips such as how to refresh the repository without having to commit files that you didn't modify due to the new interpretation of the CRLF

I agree that that is good documentation for fixing that specific issue, but it is irrelevant since `.gitattributes` is out of the scope of this lesson. For that same reason, the section on per-repository settings is also irrelevant.

As for the relevant part of the GitHub page (from the start of the page up to and including the section "Global settings for line endings"), it is essentially the same content that exists in the Carpentries lesson: an explanation of how line ending bytes get added to files (pressing Return), that different OSes do this differently, and that this is a potential problem for collaborative work, and a command to set the config option for each platform. Therefore, it is barely useful, if at all, to link to it *from the lesson* (although it may be useful in other contexts).

Instead, the Git book explains the issue on a lower level, i.e. what Git does to the bytes in the file (e.g., "auto-converting CRLF line endings into LF when you add a file to the index, and vice versa when it checks out code onto your filesystem", etc.) -- _actually_ more information about the issue. It may be terse, but I don't really see anything wrong with that, since the info there is presented as supplementary/advanced anyway ("beyond the scope of this lesson").

Anecdotally: I had never used that config option before, so I was interested in learning the justification for its use, which I understood only after reading the Git book explanation.

_episodes/02-setup.md
```diff
@@ -54,7 +54,7 @@ For these lessons, we will be interacting with [GitHub](https://github.com/) and
 > Because Git uses these characters to compare files,
 > it may cause unexpected issues when editing a file on different machines. 
 > Though it is beyond the scope of this lesson, you can read more about this issue 
-> [on this GitHub page](https://help.github.com/articles/dealing-with-line-endings/).
+> [in the Pro Git book](https://www.git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_code_core_autocrlf_code).
```
nsoranzo, line 57:
```suggestion
> [in the Pro Git book](https://www.git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_core_autocrlf).
```

markmatney: Thanks @nsoranzo

markmatney:
That sounds good to me. I will add another commit that links `_extras/guide.md` to the GitHub info.

I absolutely agree that the GitHub page is more beginner-friendly. However, I do need to point out that [at the time I opened this issue](https://web.archive.org/web/20190708190322/https://help.github.com/articles/dealing-with-line-endings), the section on refreshing a repository started with (emphasis mine):

> After you've set the core.autocrlf option **and** committed a .gitattributes file

I saw that as out of scope, since `.gitattributes` is out of scope.

munkm:
That sounds great. Thank you for adding the link to the instructor guide. 

I'm sorry it took some time to get to your PR. 

I still think the extended information on how to update a repo after line endings are included is probably too much information to have at this point in the lesson. It is very useful and has a good place in the guide.

markmatney:
No worries! Please let me know if there's any other changes I should make. Note that I also updated an outdated link in the instructor guide.

_extras/guide.md
```diff
@@ -302,7 +305,8 @@ web-hosted private repositories.
 [git-parable]: http://tom.preston-werner.com/2009/05/19/the-git-parable.html
 [github]: https://github.com/
 [github-gui]: https://git-scm.com/downloads/guis
-[github-line-endings]: https://help.github.com/articles/dealing-with-line-endings/#platform-all
+[github-line-endings]: https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings#platform-all
```
nsoranzo, line 308:
```suggestion
[github-line-endings]: https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings
```

markmatney: Thanks. I've made this change now, ready for review again.

markmatney:
Are there any other changes I should make to this PR?

munkm (approved):

munkm:
Nope! this looks good now. Thanks for checking in!

Issue #435, opened 2017-08-17
Cover how to exit VI/VIM


ErinBecker:
Originally submitted to instructor training repo. Please see discussion here: https://github.com/swcarpentry/instructor-training/issues/487

lwasser:
just note that this is covered in the config materials -- http://swcarpentry.github.io/git-novice/02-setup/#exiting-vim

iglpdc:
Yes, it covered already in the setup. Thanks anyway, @ErinBecker.

PR #681, opened 2019-07-31
Removed 'Exiting Vim' section

cypayne:
I removed the subsection titled "Exiting Vim" because: 1) it wasn't relevant to the Git Setup instructions and, because of this, 2) would likely confuse first-time Git users. This subsection instructed the student on how to exit a Vim session, however there is no mention or instance of opening Vim throughout this lesson - later in the lesson, the text editor 'nano' is used.

I kept the one piece of relevant information from this subsection - that Vim is usually Git's default editor - and appended it to the "Git editor configuration" section.

### DIFF
--- a/_episodes/02-setup.md
+++ b/_episodes/02-setup.md
@@ -96,14 +96,7 @@ Dracula also has to set his favorite text editor, following this table:
 | Vim                | `$ git config --global core.editor "vim"`   |
 | VS Code                | `$ git config --global core.editor "code --wait"`   |
 
-It is possible to reconfigure the text editor for Git whenever you want to change it.
-
-> ## Exiting Vim
->
-> Note that Vim is the default editor for many programs. If you haven't used Vim before and wish to exit a session without saving
-your changes, press <kbd>Esc</kbd> then type `:q!` and hit <kbd>Return</kbd>.
-> If you want to save your changes and quit, press <kbd>Esc</kbd> then type `:wq` and hit <kbd>Return</kbd>.
-{: .callout}
+It is possible to reconfigure the text editor for Git whenever you want to change it. Note that Vim is usually the default editor for Git.
 
 The four commands we just ran above only need to be run once: the flag `--global` tells Git
 to use the settings for every project, in your user account, on this computer.


kekoziar:
I think it's important to keep how to exit vim/vi in the lesson because it is the default text editor in some systems, whether the learner knows it or not.  While we don't teach or mention vim/vi, there may be instances when an instructor or helper needs to help a learner exit, and vim/vi is not intuitive.  The inclusion of how to exit vim originated from instructor training.  The full thread starts at https://github.com/swcarpentry/git-novice/issues/435

To paraphrase one of the comments, 
> the choice of vim/vi is made for the learner when making a commit in git if they don't specify their commit message with -m - not only is vi the default text editor in unix-like systems, it is also installed with git on windows.

cypayne:
Sure, I can see the validity in having a note about exiting vim/vi. However I still think it's misplaced - maybe it would be more relevant in the "Tracking changes" section (04), after the introduction of "commit" (i.e. "If you accidentally "commit -m" with no commit message, and you enter the vim editor, you can exit by ..."). Thoughts?

kekoziar:
I have experienced this issue with vim/vi and a novice learner during a workshop, so for me, it's not a hypothetical. There's no reason to think the absence of `-m` with a commit will happen only in Section 4. Placing the callout near the setup of the text editors creates stronger mental mapping, where people are more likely to remember "oh, the tip is near the text editors information" than "it's somewhere in the lesson" (This also happened during our workshop).

TomKellyGenetics:
I object to this change strongly. Vim is the default editor for many git installs. Unless the configuration is changed, many participants will be thrown into vim to enter commit messages. This can cause widespread confusion and inexperienced helpers may also have trouble with it.

We've experienced this several times with our workshops. Adding configuration for alternative editors and instructions on how to quit were based on extensive discussions at the time. We added it at the beginning so that participants can choose an editor or continue with vim (knowing how to use it). This way they're aware of this ahead of time and it isn't an unsettling surprise (how to exit the vim editor has literally millions of views of stack exchange for this very reason).

I'm not here to debate which is your favourite editor if you're experienced. Just to make sure that participants can focus on learning the shell or git without ending up stuck in an editor they aren't familiar with.

cypayne:
I see both of your points - apologies for the suggested edit!

munkm:
There's no need to apologize @cypayne! These discussions are always useful to have! We appreciate you thinking of ways to improve the lesson, even if it ultimately doesn't get incorporated.

cypayne:
Thanks @munkm!

Issue #671, opened 2019-06-27
Refer core.editor config to official list


katrinleinweber:
Hello :-)

In [02-setup](https://github.com/swcarpentry/git-novice/blame/04ffeff2715ba95433751f6bb0b2df978d6ed3af/_episodes/02-setup.md#L78-L91) we maintain a long list of `git config core.editor …` commands.

Because [Git's manual now has an official list](https://git-scm.com/book/en/v2/Appendix-C%3A-Git-Commands-Setup-and-Config#_core_editor), how about we reduce ours to `… "nano -w"` as [our default](https://github.com/swcarpentry/git-novice/blame/04ffeff2715ba95433751f6bb0b2df978d6ed3af/_extras/guide.md#L99-L107) example and refer to the manual?

IMHO, this is a rare chance to [reduce](https://github.com/swcarpentry/git-novice/blame/04ffeff2715ba95433751f6bb0b2df978d6ed3af/CONTRIBUTING.md#L93) the amount of content we have to maintain, without any risk, because that official list can be maintained by a wider community, with particular incentive for the editor developers themselves to keep "their" command up-to-date ;-)

kekoziar:
Hmmm. I like this, but I also worry about it. 

Git's exhaustive list isn't exhaustive (we include Notepad++ and Emacs, where they don't).

I like how we list the two different config commands for 32- vs 64-bit editors; ours is easier to cut and paste, whereas git's list requires full reading of the page (they don't even asterisk the 64-bit commands to refer to the note, so someone looking for a 32-bit command could leave the page in frustration). Which leads to...

Do we really want learners who are on the lesson page to leave the lesson pages as part of the lesson?  Granted, git-novice has plenty of links to outside references and resources; more than I've seen in other lessons.  But, I also know we are careful about outside links, especially those that don't support good pedagogy.  git-scm.com may be *the authority* on git, but they aren't exactly the best at explaining how to use git, especially for a novice learner; nor are they the best at website design.  For example, instead of the "exhaustive list" being exhaustive, the information is located in at least three different places, one of which isn't even on the same page (I now see where the Notepad++ and Emacs configuration commands are located).

So, removing the table from the lesson may condense the text, but are we really saving time in the lesson? We’re not reading through the list of text editors (or I hope no instructor sees that list and thinks to read through it), but it is there for reference, to ease setup. Removing the table in favor of this link could actually add time to the lesson.

[...]

PR #682, opened 2019-08-02
Refer core.editor config to official list

katrinleinweber:
See #671.

### DIFF
--- a/_episodes/02-setup.md
+++ b/_episodes/02-setup.md
@@ -74,27 +74,16 @@ For these lessons, we will be interacting with [GitHub](https://github.com/) and
 > $ git config --global core.autocrlf true
 > ~~~
 > {: .language-bash}
-> 
+>
 
+Dracula also has to set his favorite text editor, for example `nano`:
 
-Dracula also has to set his favorite text editor, following this table:
+> ~~~
+> $ git config --global core.editor "nano -w"
+> ~~~
+> {: .language-bash}
 
-| Editor             | Configuration command                            |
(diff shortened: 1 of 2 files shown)


kekoziar:
I will reiterate my concerns in the issue here.  There is risk of adding time to the lesson:

> So, removing the table from the lesson may condense the text, but are we really saving time in the lesson? We’re not reading through the list of text editors (or I hope no instructor sees that list and thinks to read through it), but it is there for reference, to ease setup. Removing the table in favor of this link could actually add time to the lesson.

"reduce the amount of content we have to maintain," in this case is not the same as removing material covered in the lesson. It's just redirecting instructors and learners to a site that does not display the content in as organized fashion as the lesson.

While I think a link to git's non-exhaustive table would be nice, I do not think it should replace the table in the lesson.

katrinleinweber:
> "reduce the amount of content we have to maintain," in this case is not the same as removing material covered in the lesson. It's just redirecting instructors and learners to a site

Yes, exactly, so that site's content doesn't have to be maintained here. Duplicated. 

Besides, the lesson recommends using `nano` and that's the command that my PR leaves in. What you say about "could actually add time to the lesson" depends upon how many different editors are actually used in a SwC workshop?

I would argue that everyone who uses a different editor is probably one level more advanced than our lesson assumes, by prescribing `nano`.

> that does not display the content in as organized fashion as the lesson.

It can be improved over there, for _everyone_, can't it?

kekoziar:
re #671.

PR #684, opened 2019-08-06
Add explanations for additional git log flags

### LINKED ISSUE #536: Git log example has extra options that aren't explained.
Looking at: http://swcarpentry.github.io/git-novice/04-changes/index.html

I think that adding the --graph option is useful in general, but the fact that this is not explained and does not actually show anything understandable in the output took away from when I was working through this.  

---

 ```shell
 $ git log --oneline --graph --all --decorate
 ```

```shell
 * 005937f Discuss concerns about Mars' climate for Mummy (HEAD, master)
 * 34961b1 Add concerns about effects of Mars' moons on Wolfman
 * f22b25e Start notes on Mars as a base
```

---

Can this edited to remove --graph as an option?  If it would be more informative later, we can add it later with a description.

Alternatively, lets add callouts for all the other options listed so they aren't just dropped in with no explanation.

huddlej:
Adds explanations for the flags used in the example for how to combine multiple git log flags with `--oneline` and corrects the output for the first `git log --oneline` example where the commit hashes should not have "*" prefixes (these prefixes are drawn by the `--graph` flag).

For clarity, this commit also removes the `--all` flag from the combined flags example since this flag does not change the output of git log in the simple repository of this lesson and requires a longer explanation of git references than there is time to cover during the novice lesson. The description of the `--decorate` flag does provide a link to the official documentation for git references, for learners who wish to study this subject in more detail on their own.

Closes #536.

### DIFF
--- a/_episodes/04-changes.md
+++ b/_episodes/04-changes.md
@@ -536,17 +536,20 @@ Date:   Thu Aug 22 09:51:46 2013 -0400
 > ~~~
 > {: .language-bash}
 > ~~~
-> * 005937f Discuss concerns about Mars' climate for Mummy
-> * 34961b1 Add concerns about effects of Mars' moons on Wolfman
-> * f22b25e Start notes on Mars as a base
+> 005937f Discuss concerns about Mars' climate for Mummy
+> 34961b1 Add concerns about effects of Mars' moons on Wolfman
+> f22b25e Start notes on Mars as a base
 > ~~~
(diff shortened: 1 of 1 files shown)


munkm:
The discussion in swcarpentry/git-novice#536 suggests moving the explanation to a callout. However, I think the way this PR has been constructed it makes sense in the body of the lesson.

Issue #685, opened 2019-08-14
Episode 3 suggests to use git status before initializing a repo, but does not show that


jdvorak001:
In a solution to a challenge in Episode 3 we say it would be good practice to check the status of the current directory just before issuing ``git init``. But this practice is not actually shown in the lesson.

As a side-effect, the lesson actually introduces ``git status`` without being explicit about it.
I find this a bit underhands.

jdvorak001:
And also, Episode 4 first uses ``git status`` in a context where it says "If we check the status of our project again". 
I believe we should admit that we are introducing ``git status`` in Episode 3 already.

Sipley:
I agree.  The solution to "Places to Create Git Repositories" is a little confusing because the command 'git status' doesn't give the output shown for the example as given. Suggestion: either move "if it looks like the following, you are good to go to create a new repository" earlier in the lesson when introducing 'git init' or change to "if 'git status' returns the following, it's best to avoid creating a new repository to be sure there is no conflicting repository in the directory". Alternatively, maybe combine the "Places to Create Git Repositories" and "Correcting 'git init' Mistakes" sections, emphasizing the solution for the latter & highlighting the solution for the former as a useful tip?

jdvorak001:
@Sipley, thanks for your comments. I suggested a solution to this issue in my pull request #686, which I have just updated with the latest `gh-pages` contents. Do you think the changes I suggested there could do the job, or should we try to refactor more?

Sipley:
If I understood the changes correctly, yes!  I think those are great!

kekoziar:
In preparation of the new workbench lesson infrastructure, #686 was closed with comments which I will echo here:  

PRs should be limited to one issue per PR. The submitted PR attempted to do many things, and so the original issue - introducing `git status` at the very beginning - was lost and never implemented. I happen to tell learners that if they need to remember only one thing, it's `git status`. I absolutely agree with and approve moving it to prior to `git init` with explanation. However, there were questions about the other content that was added to the PR. Please submit issues as separate PRs in the future.

Feel free to submit a solution to move `git status` prior to `git init` with explanation after the lesson has migrated to the new infrastructure.

kekoziar:
> In preparation of the new workbench lesson infrastructure, #686 was closed with comments which I will echo here:
> 
[...]

PR #686, opened 2019-08-14
Make it clear that git status is introduced in Episode 3 (issue #685)

jdvorak001:
I am suggesting this fix to issue #685.


munkm:
Hey @jdvorak001 thanks for the review request! I'm sorry it's taken a bit to get to your PR. I'll do my best to get to it this week!

kekoziar (commented):
I don't think that starting with explaining nested repositories is the best introduction to Git for new learners. It seems like they should first learn out to initialize a repository and what that means, then a little later have them do a thought experiment as to what should be made into repositories.

jdvorak001:
Thanks for the review. It certainly is not the intent of this path to explain nested repositories, let alone promote them. It just warns against them (as does the currently released version). What I tried to plant is the good practice of running `git status` before `git init`. With that, the `git status` command is introduced a little.

_episodes/03-create.md
```diff
@@ -44,21 +47,36 @@ $ cd planets
 {: .language-bash}
 
 Then we tell Git to make `planets` a [repository]({{ page.root }}{% link reference.md %}#repository)
--- a place where Git can store versions of our files:
+-- a place where Git can store versions of our files.
 
+It is good practice to first check if this is a good place to do so:
+a new repository shall not be nested in another repository placed higher up the directory hierarchy.
```
kekoziar, line 53: this sentence introduces topic of nested repositories. Learners will ask, "what does that mean" and "why shouldn't we do that" which is explained later in the lesson.

kekoziar (commented):
This PR basically moves the "Places to Create Git Repositories" exercise into the lesson prior to learners first initializing a repository.  I don't agree with the change because learners do not have the framework of initializing repositories on which to add the concept of nested repositories.

kekoziar:
I'm going to close this PR in preparation of the new workbench lesson architecture, but with the following points which halted the original merging of this PR.

PRs should be limited to one issue per PR.  This PR attempted to do many things, and so the original issue - introducing `git status` at the very beginning - was lost and never implemented.  I happen to tell learners that if they need to remember only one thing, it's `git status`. I absolutely agree with and approve moving it to prior to `git init` with explanation.  However, there were questions about the other content that was added to the PR.  Please submit issues as separate PRs in the future.

PR #690, opened 2019-09-16
Removing "#" in "git status" output demo in 03-create.md

hzhu16:
I suggest removing "#" in "git status" output demo to match the actual output on the terminal.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -88,10 +88,10 @@ $ git status
 ~~~
 {: .language-bash}
 ~~~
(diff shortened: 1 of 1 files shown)


peterjc:
I wonder if older versions of git used hashes?

Anyway, there are more cases to fix - eg in 06-ignore

munkm:
This is a good catch! Thanks @hzhu16! I've opened another issue to notify future contributors that they may want to fix other instances of the issue you found here. Thank you for this contribution! 🎉

PR #699, opened 2019-11-13
Update log --graph output & remove redundant --decorate

katrinleinweber:
I'm not exactly sure [when Git changed this](https://github.com/git/git/search?l=Text&q=log+graph+decorate), but I checked these changes in both Windows' Git bash (2.22.0.windows.1) & on macOS' Git (2.24.0): `--decorate` is no longer necesary to show `HEAD` & branch names. Thus, that concept could be remove to simplify the lesson.

- [x] Can someone please verify on Linux?

### DIFF
--- a/_episodes/04-changes.md
+++ b/_episodes/04-changes.md
@@ -553,7 +553,7 @@ Date:   Thu Aug 22 09:51:46 2013 -0400
 > ~~~
 > {: .language-bash}
 > ~~~
-> * 005937f Discuss concerns about Mars' climate for Mummy (HEAD, master)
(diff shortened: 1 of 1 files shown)


nhejazi:
From a quick check, this seems to be the case, though I hadn't realized this update to git had happened. I check on my local Ubuntu 18.04 installation:
```bash
$ uname -v
#75-Ubuntu SMP Tue Oct 1 05:24:09 UTC 2019

$ git --version
git version 2.17.1
```

Since the `--decorate` flag appears to be superfluous now, I agree that it should be remove to simplify the lesson. Merging this accordingly --- thanks very much for pointing this out!

katrinleinweber:
You're welcome & thanks for merging :-)

PR #700, opened 2019-11-13
added typical number of characters when using short commit ids

modenaxe:
The original document did not specify how many character you should use when using short ids for identifying commits. The [official documentation](https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection#Short-SHA-1) page suggests eight to ten characters, but from the context it is clear that for "normal" size projects seven characters are also good. 
Seven characters are also consistent with the "tracking changes" lecture where the command:
```
git log --oneline
```
is used and outputs logs with 7-characters ids for commits.

### DIFF
--- a/_episodes/05-history.md
+++ b/_episodes/05-history.md
@@ -156,7 +156,7 @@ index df0654a..93a3e13 100644
 
 That's the right answer,
 but typing out random 40-character strings is annoying,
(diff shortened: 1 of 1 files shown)


nhejazi:
This looks good to me. Merging since it's a small change. Thanks for your contribution.

PR #702, opened 2019-11-18
Remove # from output, because not printed by git

katrinleinweber:
These `#` may have been present years ago, but are no longer.


katrinleinweber:
You're welcome & thanks for merging :-)

PR #705, opened 2019-12-03
Update objectives to include learning about the .git folder

vyasr:
Update the learning objective for the git folder.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -6,6 +6,7 @@ questions:
 - "Where does Git store information?"
 objectives:
 - "Create a local Git repository."
+- "Understand the purpose of the `.git` directory."
 keypoints:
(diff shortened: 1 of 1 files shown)


kekoziar:
I'm going back to my original comment on learning objectives.  Learning objectives need to be measurable and stated in an active voice. You can't measure "to learn" or "understand." But, you can measure if someone is able to perform something or how accurately they describe it.  This is one of the tenets of [developing and delivering Carpentries content](https://carpentries.github.io/instructor-training/15-lesson-study/index.html).

I think the proposed objective should be changed to *"Describe the purpose of the .git directory"* which is in line with the framework Bloom's Taxonomy provides for learning objectives.

vyasr:
I agree and have updated to reflect this change.

kekoziar:
Thanks!

PR #706, opened 2019-12-03
Reword section describing git init

vyasr:
Improves and fixing spacing in the description of `git init`.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -47,12 +47,12 @@ $ git init
 ~~~
 {: .language-bash}
 
-It is important to note that `git init` will create a repository that
-includes subdirectories and their files---there is no need to create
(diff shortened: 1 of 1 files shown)


kekoziar:
Thanks for the change.  I'm not sure I agree it's needed. It could be that the change moves the tone of the beginning of the section from conversational to written, and so it's throwing off the rest. Since this curriculum is sometimes read verbatim by new instructors, I think it's important to keep in mind how it sounds if you're *speaking* it to someone who doesn't know git.

vyasr:
I think that's fair. In general, I always prefer writing "Note that..." to "It's important to note that..." since stating importance seems redundant. However, the latter phrase feels more conversational. I'm fine rejecting this PR if you'd prefer.

PR #707, opened 2019-12-03
Apply minor edits to text in 04

vyasr:
Improve description of how staging changes works in `git`.


kekoziar:
Looks good.  I like this change because it moves the language from something you might need to explain (*snapshot* is a dated term for a photo, and can mean a lot of things now) to more explicit language.  Nice change.  Thanks!

Issue #519, opened 2018-05-12
This description of the VCS concept is misleading for Git


agzimmerman:
[This description](https://github.com/swcarpentry/git-novice/blob/gh-pages/_episodes/01-basics.md) of the version control concept is misleading for Git. Since this is specifically in the git-novice lesson, I think it should be changed.

In the case of Git, rather than how Subversion works, when you commit, **Git stores entire copies of any (staged) files that were changed**. It does **not** recreate versions by applying sequences of diffs. This is explained [here](https://stackoverflow.com/questions/8198105/how-does-git-store-files).

iglpdc:
There is some related discussion [here](https://github.com/swcarpentry/git-novice/issues/263).

The point is that this episode is more an introduction to version control systems rather than to Git. (I would be ok to change the second objective to reflect that). The way Git stores things internally is pretty fascinating, but at this level maybe it's an implementation detail. We recommend instructors to read and know how Git works (see the instructor's guide), but for learners might be confusing.

I think that trying to explain what is really happening would either complicate things to much for some learners, or would take much more time.

I'm closing this for now, but feel free to reopen if you can think of a modification that can both be more technically accurate without adding much time or extra complications to the lesson.

I'm also opening a new issue with the change of the second objective.

alexanderzimmerman:
Thanks for considering this.

I do think this issue should still be open; but I don't see that I have an option to re-open it. How do I re-open the issue?

I agree entirely about keeping things simple; and, as you suggest, I see no need to teach the graph model discussed in #263. The good news is that **we can teach a more accurate mental model** and, at least to me, it's **even more simple** than the diff mental model we're teaching. Maybe if you've been using Subversion for years, then it's hard to let go of this idea of applying a sequence of diffs to recover different versions. But **being naive, it's much easier to picture storing copies of any files that are changed. This is actually closer to what we do manually without version control systems.**

Also for reference, in [this later episode](https://swcarpentry.github.io/git-novice/04-changes/index.html) , it would help to already have a correct mental model for how git stores changes, especially when we say "When we run git commit, Git takes everything we have told it to save by using git add and stores a copy permanently inside the special .git directory." It was actually when I was reviewing this lesson for the first time that I looked up how git actually stores changes, since I had been assuming the Subversion diff concept.

Thoughts?

iglpdc:
Ok, we can reopen and let others comment.

[...]

PR #715, opened 2020-01-08
Update the objectives in episode 1

tkoskela:
Updated the second objective in episode 1 to solve issue #519. I like the original suggestion in the issue since it's more consistent with the first objective, and sounds less complex to me. At this point the student really only needs to very broadly understand what a version control system does, not so much what operations are performed.


munkm:
Note: The suggestions in swcarpentry/git-novice#519 are not all implemented yet, so I'm leaving that issue open.

PR #717, opened 2020-01-14
Correct error on counting commits in episode 05

gefux:
Hey everyone! 
Following the course step by step one finds that HEAD~3 doesn't exist and 
`$git diff HEAD~3` would in fact return:
```
fatal: ambiguous argument 'HEAD~3': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

The output that we see in the lesson is achieved with `$git diff HEAD~2` instead.

### DIFF
--- a/_episodes/05-history.md
+++ b/_episodes/05-history.md
@@ -74,7 +74,7 @@ again, but with the notation `HEAD~1`, `HEAD~2`, and so on, to refer to them:
 
 
 ~~~
-$ git diff HEAD~3 mars.txt
+$ git diff HEAD~2 mars.txt
 ~~~
 {: .language-bash}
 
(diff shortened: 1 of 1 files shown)


kekoziar:
The output of the lesson includes callouts (which are important caveats, and part of the main lesson) but not exercises. Episode 4 has a callout on directories, so the fourth commit is `git commit -m "Add some initial thoughts on spaceships"`

gefux:
Oh, yes, I see, my bad! Thanks though!

tboby:
@kekoziar The Curriculum Development Handbook states
> {: .callout} for sharing an aside or comment. Use sparingly.

https://cdh.carpentries.org/technological-introductions.html#using-the-lesson-template

And the content of many of the callouts in the previous episodes seem irrelevant to most target audiences (proxies? exiting vim when we explicitly recommend nano?).  They'd also considerably push the episode times over the stated limits!

Is this expectation (that callouts are not optional, niche, asides) documented anywhere? If not, should this example be tailored to cover both options somehow?

Issue #718, opened 2020-01-15
Episode 14 - Challenge solution should be in R


jrkirk61:
While reviewing episode 14, which is the supplemental Rstudio lesson, I noticed that the answer to the [challenge question ](http://swcarpentry.github.io/git-novice/14-supplemental-rstudio/index.html) is given in bash code. Wouldn't it make more sense to give the answer in R, since the lesson is about using git in Rstudio and it can be assumed that attendees would have some experience with it if they are interested in this lesson? When I was trying to guess the answer myself, I immediately thought to try and answer it within Rstudio and was surprised to see that the answer was in bash. 

If you agree, below would be my proposed changes to the Challenge solution:

```
> dir.create("./graphs/")
# Open .gitignore file in the right-hand panel, add "graphs/" to the list and then save.
```

munkm:
Hi @jrkirk61! Thanks for bringing this up and offering a suggestion for an alternative solution. I think it's a great idea to have a solution here that's done completely in the RStudio GUI. It's more consistent with this episode and it will reinforce some RStudio principles. Would you be interested in submitting a PR to update the solution to what you've proposed?

jrkirk61:
Done!

munkm:
Awesome @jrkirk61! I see you've committed this change to your fork of the repo. Would you mind turning it into a pull request so we can use your changes in the main lesson here in swcarpentry? If you're on https://github.com/swcarpentry/git-novice or on your fork at  https://github.com/jrkirk/git-novice you should see a banner saying that your branch `patch-1` differs from the `gh-pages` in this repository and directs you to make a PR. Let me know if you need any help or don't see it or something! We'd love to see your contribution here too!

jrkirk61:
I think I submitted a PR? Haha. I've never tried to do a PR before, so let me know if it's still not right.

munkm:
You definitely submitted one! 🎉 Congratulations! I'll review it shortly. Thank you so much for contributing!

kekoziar:
Issue resolved.

PR #719, opened 2020-01-15
Make challenge solution Rstudio friendly

jrkirk61:
Please see issue #718 for the reasoning of this request.


munkm:
This looks great @jrkirk61! Thank you so much for the RStudio solution! I'm unfamiliar with this tool so it's really great to have your perspective. Thank you.

PR #729, opened 2020-03-11
Remove confusing `rm` text in episode 3

dbrower:
The challenge problem to remove a nested git repository at the end of episode 3 has a confusing solution. It uses the term "working tree" that is not used anywhere else in the course. It discusses "working directory" before that is introduced in episode 4. Furthermore, removing the errant `.git` directory is not the same as "removing files from the repository", and in general an `rm` command will not remove files from the history of a git repository.

Others have also noticed this problem. Among all the changes in #686 there is a similar edit.

### DIFF
--- a/_episodes/03-create.md
+++ b/_episodes/03-create.md
@@ -156,26 +156,18 @@ wording of the output might be slightly different.
 >
 > > ## Solution -- USE WITH CAUTION!
 > >
-> > ### Background
-> > Removing files from a git repository needs to be done with caution. To remove files from the working tree and not from your working directory, use
-> > ~~~
-> > $ rm filename
-> > ~~~
-> > {: .language-bash}
-> > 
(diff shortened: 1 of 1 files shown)


kekoziar:
The original PR and issue indicates a concern of providing a very powerful and potentially dangerous command w/o background.   
https://github.com/swcarpentry/git-novice/pull/588, https://github.com/swcarpentry/git-novice/issues/558
I agree that the prefaced text needs to be clarified since working tree and working directory can be confusing.  Feel free to modify your PR to keep the needed background (not everyone will have gone through the shell lesson), while updating the language.

kekoziar (changes requested):
I agree that the prefaced text needs to be clarified since working tree and working directory can be confusing. Feel free to modify your PR to keep the needed background (not everyone will have gone through the shell lesson), while updating the language

kekoziar:
resolved in https://github.com/swcarpentry/git-novice/pull/753

PR #732, opened 2020-04-24
Add link to Creative Commons FAQ in instructor notes

### LINKED ISSUE #731: Applying a Creative Commons license to software
One of the Key Points in the Licensing episode is
> The Creative Commons family of licenses allow people to mix and match requirements and restrictions on attribution, creation of derivative works, further sharing, and commercialization.

I've found that using Creative Commons licenses for software is a common source of confusion.  Would it be useful to link to the Creative Commons [FAQ](https://creativecommons.org/faq/#can-i-apply-a-creative-commons-license-to-software) about this?  Perhaps the instructor notes would be best so that it doesn't clutter the Key Points.

agitter:
Closes #731

This addition to the instructor notes clarifies the Key Point about Creative Commons licenses and can guide discussion of software licenses.

### DIFF
--- a/_extras/guide.md
+++ b/_extras/guide.md
@@ -285,6 +285,10 @@ what, arise naturally once we start talking about using public services like
 GitHub to store files. Also, the discussion gives learners a chance to catch
 their breath after what is often a frustrating couple of hours.
 
+The Creative Commons family of licenses is recommended for many types of
+works but not software. Creative Commons [recommends][cc-faq-software]
+a software-specific license instead.
+
 ## [Citation]({{ page.root }}{% link _episodes/12-citation.md %})
 
 ## [Hosting]({{ page.root }}{% link _episodes/13-hosting.md %})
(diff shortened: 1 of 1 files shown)


_extras/guide.md
```diff
@@ -285,6 +285,10 @@ what, arise naturally once we start talking about using public services like
 GitHub to store files. Also, the discussion gives learners a chance to catch
 their breath after what is often a frustrating couple of hours.
 
+The Creative Commons family of licenses is recommended for many types of
+works but not software. Creative Commons [recommends][cc-faq-software]
```
munkm, line 289: perhaps add something like "many types of works (including project documentation) but not software", so that learners can understand the context of learning about CC licenses in a software workshop.

agitter: Good idea.  I updated this in 0e2fdee6d24ddaaae3b5fbb233ce233fa731f5b4.

munkm:
Awesome! Thanks for adding this!!! 🎉

PR #737, opened 2020-05-22
Fixed the link redirecting to the Git setup page

xdebbie:
Link was broken, this edit fixes it and redirects the user to the correct Git setup instructions page.

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -14,4 +14,4 @@ $ cd Desktop
 ~~~
 {: .language-bash}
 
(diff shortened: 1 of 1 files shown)


munkm:
This is a good interim solution. I'd like to note that we should find a longer term solution for linking to rendered installation directions (I think this will not be straightforward for all people to parse). Once we have a resolution to carpentries/workshop-template#660 then we can do another update. 

Thank you so much for your PR @xdebbie

PR #742, opened 2020-06-09
episodes: proofread changes for instructor checkout

rmcdermo:
Thanks for the opportunity to contribute!

In this PR:
* Please consider updating the title of the first section of the lesson from "Automated Version Control" to "Explicit Version Control" (or similar).  In my view, Git is not automated (that would be a bot).  The commits we make are deliberate and manual.
* I added comments to sections of the lessons where I felt that `fetch` and `merge` should be discussed in connection with `pull`.  I'd be happy to contribute more to the lesson if the maintainers agree.  It seems that there has been a deliberate decision made to use `pull`, perhaps for simplicity.  But this becomes problematic later when we discuss conflicts, for example.
* I added a section on the use of `Personal Access Tokens` to the `github.md` for pushing commits if two-factor authentication is in use.
* Typos and consistency of comma after "e.g." and "i.e.".
* My editor cleans all white space on saving the file, so you will see several lines with only this change.

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -14,4 +14,4 @@ $ cd Desktop
 ~~~
 {: .language-bash}
 
-[workshop-setup]: https://github.com/carpentries/workshop-template/blob/gh-pages/_includes/swc/setup.html
+[workshop-setup]: https://carpentries.github.io/workshop-template/#git


grantmeadors (commented):
Reviewing changes for pull request #742 by rmcdermo. Please see line-by-line comments, as well as general comments below. Overall, the pull request makes a substantial number of improvements to the grammar and spelling of the lesson.

Additional review is required for some elements of the request.

The decision whether or not to discuss pull as a combination of fetch and merge needs to be considered carefully. It is true that fetch and merge are more fundamental concepts, but many novice git users are able to begin using git while only using pull. The current lesson is not, in my opinion, misleading by omitting fetch and merge; the simplicity of pull may increase comprehensibility. However, users can easily run into merge conflicts on multi-user (or even multi-computer) projects, which they will not know how to resolve without having encountered merge.

Finally, the two-factor authentication (2FA)/HTTPS login token discussion is important. Users increasingly rely on 2FA, and the lesson must be able to accommodate them. However, if an extended diversion is needed to do so, perhaps it would be better to rework the section to emphasize more-secure SSH authentication instead.

Other than those two issues, the pull request looks like an improvement, and I recommend its inclusion.

PR #760, opened 2020-10-06
Change `checkout` actions to use newer `restore` syntax

### LINKED ISSUE #691: Consider replacing "checkout" with "switch" and "restore", if widely available in the future
`git checkout` has been confusing for learners (especially when they later discover branching), see https://github.com/swcarpentry/git-novice/issues/169 for a first episode of this discussion.

This summer 2019, git started an experimental feature to solve this issue replacing it with two commands `git restore` and `git switch`
https://github.blog/2019-08-16-highlights-from-git-2-23/#experimental-alternatives-for-git-checkout

I though this was interesting to share, if people want to try the feature and/or push for its adoptions and/or start thinking about how the lessons might need to be modified.

Cheers.

jat255:
This PR resolves #691 and changes all references to `git checkout` to either `git restore` (or `git switch`) as appropriate. I believe I got all the references, but a close look at the changes my a maintainer would be great. Since 2.28.0 is the default version in newer versions of Git for Windows and Linux, I think it makes sense to make these changes now. I've included the older syntax in comments in case people are using an older version, too.

I also tried to update the output blocks to reflect the verbiage used by git in more modern versions (I ran all the examples with Git 2.28.0 on Linux).

### DIFF
--- a/_episodes/04-changes.md
+++ b/_episodes/04-changes.md
@@ -282,8 +282,9 @@ If we break it down into pieces:
 2.  The second line tells exactly which versions of the file
     Git is comparing;
     `df0654a` and `315bf3a` are unique computer-generated labels for those versions.
-3.  The `i` and `w` (your version might say `a` and `b` if using an
-    older version of Git) refer to the `index` and `working` copies
+3.  The `i` and `w` (your version might say `a` and `b`
+    depending on settings; see note below) refer to the `index`
+    and `working` copies
     of the files. The "working" version is the one that we have changed,
     while the "index" is the version that Git already knows about.
 4.  The third and fourth lines once again show the name of the file being changed.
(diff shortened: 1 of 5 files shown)


kekoziar (changes requested):
Hi, @jat255.  Thanks for working on the needed change that updates the listed output.  However, please remove the changes to the identifiers and dates.  Also, I don't think exposition to differences between git versions is needed (nor usually incorporated in the lesson).  Items like i/w and mnemonicPrefix were never introduced into the lesson, and are out of scope.

I do appreciate the changes from `checkout` to `restore` and `directory` to `tree`

matuskalas:
I think the time is now very due to finally retire `checkout` from the lessons, and replace with `restore` and `switch`. 😊

kekoziar:
I'm going to close this PR in preparation of the new workbench lesson infrastructure, but with the following points which halted the original merging of this PR.

Yes, the output text has changed so this can be updated; however, the original PR has many conflicts which need to be resolved, which halted being able to update and merge.  

This content should be updated, at least during the next lesson content review, if not earlier.

PR #767, opened 2020-11-05
Add note about new git pull message to instructor guide

lo5an:
Recent versions of git introduce a warning message when you run `git pull` without specifying how to reconcile divergent branches. A discussion of the options probably exceeds what we can do in a workshop, but instructors  may want to show learners how to suppress this potentially distracting warning.


kekoziar:
@lo5an  will you verify that this message is still produced, and under what conditions? Since the lesson does not delve into branches, it seems out of scope. Also, I cannot produce the message.

kekoziar (commented):
unable to produce warning message.

lo5an:
It looks like git was [updated](https://github.com/git/git/commit/c525de335e4a3919d49a9126b76d1288f35737be) at the end of 2020 so that the message isn't displayed when a fast-forward merge is possible. I'm seeing this behavior now on my workstation.

People will still see the message in non fast-forward merge situations:

kekoziar:
If you don't mind, I'm going to put a "wait" tag on this.  It looks like there are still [PRs and discussion](https://github.com/git/git/search?q=pull.rebase%3Dfalse&type=issues) about this and the various `git pull` options over at the Git repo.  

As an aside, I'm running Git version 2.32.0.windows.2 and am still not seeing the behavior.  When I dug into the different `git config --list` options, I found `pull.rebase=false` configured at the system level, meaning Git-bash or Git automatically added the configuration.

lo5an:
Sounds good. It's definitely an edge case, especially if Git for Windows is configuring the option at the system level.

lexnederbragt:
Last month I updated my Mac to Big Sur and had to reinstall git. I am now at version `git version 2.30.1 (Apple Git-130)`. When I did my first `git pull` on a repo that was already up to date, I got the message:

```
$ git pull
hint: Pulling without specifying how to reconcile divergent branches is
hint: discouraged. You can squelch this message by running one of the following
hint: commands sometime before your next pull:
hint:
hint:   git config pull.rebase false  # merge (the default strategy)
hint:   git config pull.rebase true   # rebase
hint:   git config pull.ff only       # fast-forward only
hint:
hint: You can replace "git config" with "git config --global" to set a default
hint: preference for all repositories. You can also pass --rebase, --no-rebase,
hint: or --ff-only on the command line to override the configured default per
hint: invocation.
Already up to date.
```

This _could_ imply that some learners with recent versions of git get this message. It may be worth checking with folks that have recently taught this lesson.

As an aside: is the default strategy `git config pull.rebase false` the most sound one?

kekoziar:
I'm going to close this PR in preparation of the new workbench lesson infrastructure.  I will leave the label in place, so it is easier to get back to the original solution when revisited.

Issue #764, opened 2020-10-11
Figure caption ~/vlad/planets is confusing


zingale:
For lesson 7 (http://swcarpentry.github.io/git-novice/07-github/index.html), the figure shows

http://swcarpentry.github.io/git-novice/fig/git-freshly-made-github-repo.svg

but I don't think we've referenced `~/vlad/planets/` before.  Should this be `~/Desktop/planets/`?

kekoziar:
@zingale  `~/vlad/planets` sets up the images for the next episode, where the images need to differentiate between two local computers: `~/vlad/planets` and `~/wolfman/planets`

kekoziar:
I will restate my concern:  I'm not sure I agree with this change.  There is a need to differentiate between two local computers and one remote. `~/vlad/planets` removes ambiguity of only stating `Desktop` (whose desktop?), while setting up the images for the next episode, where the images need to differentiate between the local computers: `~/vlad/planets` and `~/wolfman/planets`

I will leave this open for a short time for feedback.

kekoziar:
The proposed change introduces ambiguity and conflict for the next episode.  Items are labeled with Vlad or Wolfman as part of the story and to differentiate between computers.

PR #770, opened 2020-11-30
changed name to Desktop

JensWehner:
fixes https://github.com/swcarpentry/git-novice/issues/764


kekoziar (commented):
I'm not sure I agree with this change.  There is a need to differentiate between two local computer and one remote. `~/vlad/planets` removes ambiguity of only stating `Desktop` (whose desktop?), while setting up the images for the next episode, where the images need to differentiate between the local computers: `~/vlad/planets` and `~/wolfman/planets`

I will leave this open for a short time for feedback.

kekoziar:
The proposed change introduces ambiguity and conflict for the next episode.  Items are labeled with Vlad or Wolfman as part of the story and to differentiate between computers.

Issue #778, opened 2021-02-01
Password authentication for GitHub being phased out


edbennett:
I was delivering this lesson on Friday and shortly after introducing Remotes in GitHub I got an email from GitHub pointing out that [password authentication for command-line operations is being phased out in August](https://github.blog/2020-12-15-token-authentication-requirements-for-git-operations/), at which point all users will either need SSH keys or personal access tokens, with the latter being "recommended" (for reasons as yet unclear to me).

Previously the lesson avoided teaching SSH keys as it introduces another concept that needs to be taught, and unnecessary cognitive load when trying to simultaneously deal with "what is SSH, a public key, a private key" at the same time as they are dealing with "what is a remote, what is push", whereas entering a username and password is likely to be more familiar. Unfortunately it seems that we no longer have that luxury, so will need to redevelop the "Remotes in GitHub" section to include setting up either a key or an access token, and explain the security implications of this.

aragilar:
Given we're not using any GitHub specific features in the lesson (as written), maybe we should look at alternate hosts (e.g. GitLab, Bitbucket)? Having supplemental lessons covering setting up ssh would be useful, but given the current difficulties getting learners set up, I'm not sure staying with GitHub is worth the effort of additionally setting up ssh keys.

dokempf:
I think that the additional effort of explaining how to generate a Personal Access Token for the course work is small enough. I will prepare a PR that can serve as a basis of discussion - I had already planned to do this for my checkout anyway before @edbennett first brought up the issue.

edbennett:
I suspect that now GitHub has set a direction of travel, GitLab and BitBucket will eventually follow, so changing platforms just to avoid introducing this material would end up creating extra work.

@dokempf Thanks for agreeing to work on this!

dokempf:
So, I have added a PR that solves this issue by going the route of explaining how to generate a personal access token. I personally feel that this is the way to go, because it does not add a lot of cognitive load compared with setting up SSH. I would not switch hosting provider - after all, their reasons for doing this transitions are quite valid! Looking forward for your feedback!

jttkim:
Just a quick idea -- would it make any sense to mention / introduce ssh in the shell lesson? There's already some mention of using shells for working with clusters or other remote systems in there, and needing / wanting to work with remote systems is a relatively common reason for scientists to learn using shells, in my experience.

munkm:
[...]

PR #779, opened 2021-02-02
Explain the use of GitHub personal access tokens in the remotes episode

dokempf:
Github has deprecated password authentication from the command line and will disable it in August 2021: https://github.blog/2020-12-15-token-authentication-requirements-for-git-operations

The corresponding Carpentry GitHub issue is https://github.com/swcarpentry/git-novice/issues/778

This PR implements a solution to the problem that explains how to generate a personal access token
through the GitHub UI instead. While this method adds an additional step (including several screenshots
from the GitHub UI) to the "Remotes in GitHub" episode, IMHO it does not produce a lot of additional
cognitive load - especially when comparing to the alternative of setting up SSH during the lesson.

The PR also adds some instructor notes on how to handle tokens in live coding.

The token in the screenshot is revoked, no worries.

NB: I submit this as part of my checkout, so it is a first contribution. I might have missed some markup
styling conventions, but I am happy to incorporate any feedback that you give.


dokempf:
I am aware that this contribution is in violation of the *Do not add stuff unless you also remove stuff*-Rule. I still think this is a valid contribution for my checkout procedure, because it solves an issue that will make the lesson stop working by August 2021. If I was to decide what to remove from the episode instead, I would move the following parts into extra materials:

* Proxies
* Password managers
* Uploading files directly in GitHub browser

_episodes/07-github.md
```diff
@@ -120,8 +120,46 @@ origin   https://github.com/vlad/planets.git (fetch)
...
+
+You can now configure which operations are possible with the new personal access
+token that we are about to generate. For this course, we only need access to
+public repositories:
+
+![Configure Personal Access Token on GitHub](../fig/github-new-token.png)
+
+Scrolling down and clicking on "Generate Token" will finalize the generation process.
+Your token will be displayed on the screen. You need to copy and store this token now
+just like you normally store passwords. Note that GitHub will never again display this
```
tobyhodges, line 156: I would really like to see a link added here, pointing to a useful guide about good practices in storing passwords like this locally: it is likely that this is the first time many learners will be encountering access tokens (e.g. rather than "memorable" passwords or secure passwords stored by their browser/LastPass etc). It would be beneficial for them to be able to read about some secure ways they can use the PAT without needing to store it in plain text. @dokempf do you know of any such resources that you could link out to here?

jttkim: I use `ssh` keys, as described in the GitHub docs [1], rather than access tokens. That seems less complex to me, and it's no less secure. So would it make sense to guide learners to that solution?

If learners aren't using a password manager already, I have reservations of suggesting to them to start using one as part of a workshop lesson, when they're not prepared for that. Adopting a password manager is a decision with long-term consequences, users are to quite some extent "locked in" once they've put any appreciable number of access credentials into such a system. I'm concerned that explaining that in the minimally required detail to enable learners to make a properly informed decision will considerably disrupt the flow of the lesson (and suggesting to learners that they just adopt the thing now, without that basis, isn't desirable for this matter either).

[1] https://docs.github.com/en/github/authenticating-to-github/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent

aragilar: I think the use of ssh keys vs tokens depends on which client you use: using ssh keys with the cli git client makes sense, but what does R-Studio allow (or the variety of other clients out there)? I wonder what clients people teach with (I know when Git is taught in R workshops at MQ, it's taught via R-Studio), and whether it would be better to split up the lesson in such a way that different clients go through the concepts in a more client-focused way (similar to https://mq-software-carpentry.github.io/R-git-for-research/setup.html, where git is more integrated into the overall course).

I would say password managers shouldn't be needed (in this case): the client should be handling the security of the token (and getting that set up would be better than spending time on a long aside), which for the normal cli client is covered by https://stackoverflow.com/a/32470658/1306020.

Additionally, I would explicitly avoid comparing the token to a password, the token should be copied to where it needs to be and forgotten, if there's a need to put it somewhere else, a new token should be created.

tobyhodges: > I use ssh keys, as described in the GitHub docs [1], rather than access tokens. That seems less complex to me, and it's no less secure. So would it make sense to guide learners to that solution?

I use SSH to access GitHub too. But - in my opinion - there are more new concepts that need to be introduced for learners to understand what is happening when a key pair is created and SSH access is configured, compared to setting up an access token. It is almost certainly more time-consuming, in a lesson that is already very full. 

There is also, I think, some additional complexity for Windows users who would like to generate SSH keys?

> I would say password managers shouldn't be needed (in this case): the client should be handling the security of the token (and getting that set up would be better than spending time on a long aside), which for the normal cli client is covered by https://stackoverflow.com/a/32470658/1306020.
> 
> Additionally, I would explicitly avoid comparing the token to a password, the token should be copied to where it needs to be and forgotten, if there's a need to put it somewhere else, a new token should be created.

Right. But then a link out to that StackOverflow thread, or at least some discussion of the appropriate way to store the token, should be included in the lesson here?

aragilar: Sure, I agree with the discussion about how to handle the token (or whatever secret system is used) should be included, I'm only objecting to the use of password managers (as it's the wrong tool in this case, a keychain/local secrets manager should be used). I would not link to the stack overflow, rather move the contents of the link to the setup.

It's also worth flagging some applications use OAuth, which is a third mechanism. I think having an episode per client on how to connect up to GitHub (or move it to the setup) would have a better flow (as already exists for R-Studio), as this is a per-system thing, rather than a per-repo thing.

dokempf: I agree with all of above discussion, that a good recommendation of how to handle the token is needed, but I fail to see a good, reasonably universal solution. When originally writing this, I was not thinking outside of my password manager usage bubble.

_extras/guide.md
```diff
@@ -167,6 +167,26 @@ particular set of files in `.gitignore`.
     repositories in the web and provides a web interface to interact with repos
     they host.
 
+*   Using personal access tokens to authenticate with GitHub from the command line
+    might be confusing for some learners as GitHub has only recently deprecated
+    password authentication. You can point advanced learners interested in the
+    rationale behind the decision by GitHub to [this blog post explaining it][github-access-tokens].
```
dpshelio, line 173: `this blog post explaining it` is not a "good" link in terms of accessibility (see [What's wrong with using "click here" links](https://webaccess.berkeley.edu/ask-pecan/click-here)). I'd suggest something like the following:
```suggestion
    password authentication. You can point advanced learners interested to [the
    rationale behind the decision of deprecating passwords by GitHub][github-access-tokens].
```

_extras/guide.md
```diff
@@ -167,6 +167,26 @@ particular set of files in `.gitignore`.
...
+    to your GitHub account by using it. There is a few good strategies to address this
+    issue though:
+
+    *   Immediately revoke the token that you generated after you showed everybody
+        how generating the token works. For the rest of the course, work with a token
+        that you have generated without learners watching (e.g. before the course)
+
+    *   Be even more prudent about this in online teaching, where learners are able
+        to take screenshots.
+
```
dpshelio, line 186: I believe a "safer" suggestion would be to use a new account or one only for teaching. That's what I normally do so I know I'm seeing something similar to what the learners are seeing and it doesn't contain any information to other projects/notifications I may be working on.  Certainly, we should follow the practices you suggested here, but in case I forget to revoke that token I'm not in big trouble.

dpshelio (approved):
Great work! 💪 

Besides the small suggestions I added within the text, I'd only add to take a new screenshot of the `github-dev-settings` where your name and org is replaced, maybe by `Vlad Dracula` for your name and `Transilvan.ia` for the company as in the rest of the lesson. You don't need to save these, only to change them for the photo ;)

_episodes/07-github.md
mikerenfro: At the moment, I'm on Team SSH, though this is somewhat influenced by my HPC-centric focus, and having only done the Git lessons as the second part of a standard 2-day SWC workshop. Lightly edited copy of comments I made on the hoc-carpentry Slack channel below:

> I'm not sure that all the screenshots and steps to make a PAT and use it on each operation is an improvement of `ssh-keygen ; cat ~/.ssh/id_rsa.pub` and pasting the resulting public key into one GitHub form, especially for an audience that's already at the command line for the rest of the workshop. Even less so if they've already been through the Shell lesson as would be common.
>
> `ssh-agent` and `ssh-add`? Sure, that starts getting more wordy and complicated in trade for having to type fewer passwords/passphrases over time. But if we're looking for something no more onerous than sending a password over https, a passphrase-enabled SSH key doesn't seem to be any worse than the original setup.
>
> Downside would be for anyone on a Windows platform that doesn't have native ssh. Which appears to be any Windows version prior to the April 2018 release of Windows 10.

Microsoft pushes pretty heavily for Windows 10 users to be on the latest major release, but I couldn't quickly find the relative share of Windows 10 releases earlier than the April 2018 one. Windows 10 does appear to have about a 75-80% share of the Windows install base overall, though that can vary by region.

bkmgit:
The contribution is welcome. However access tokens are GitHub specific, whereas the aim of the lesson is to teach people to be able to use Git repositories in general. As such SSH keys are preferable. It may be helpful to have this as optional supplementary material with SSH key as the default.

kekoziar:
Hello, @dokempf.  Thank you for your submission.  The main lesson will demonstrate the minimum amount of SSH to authenticate an account, since SSH is used by many applications.  This keeps in line with The Carpentries' values to support open source solutions over proprietary (i.e. GitHub) solutions.  However, there will be a supplemental episode that covers SSH in more depth; perhaps this can go into that episode.

dokempf:
@kekoziar Don't worry, I have no hard feelings regarding this contribution. I knew exactly that there would be a discussion of SSH vs. PAT. I hope this PR was helpful for that discussion to see how explaining PATs would look like and where the potential caveats would be.

kekoziar:
Thanks, @dokempf.  I'm going to close this now, but I think it should be reviewed when the supplemental episode is developed.

PR #786, opened 2021-03-15
Update alt-text for RStudio screenshot

zkamvar:
Note that the gitignore alt-text is a bit long. It can be shortened if you want to describe the files that are added by default in the text.

This contribution is part of the Core Team accessibility hackathon to add alt-text to images across our lessons. Please feel free to improve upon these suggestions as desired!


munkm (approved):
I like how verbose the alt text is here. If there are arguments for shortening it I think we can welcome them in a follow up PR.

PR #788, opened 2021-03-22
Renamed references for "master" branch to "main" branch

JonKing93:
This pull request changes all references to the "master" branch to the "main" branch instead.

Use of a "master" branch is generally discouraged as culturally insensitive and the most recent versions of git default to a "main" branch instead. Although some Carpentries users may have older versions of git that still default to a "master" branch, I think it would be best for the language on the website to reflect the more culturally sensitive (and most up-to-date) use.


kekoziar:
Multiple PRs were submitted covering the same change.  I chose the PR which was more comprehensive.

PR #789, opened 2021-03-31
Update 02-setup.md

annajiat:
Added notepad for windows as one of the option for editors.
Though only notepad should suffice, included full path just in case the environment's path variable is modified/corrupt.


kekoziar:
Confirmed.  Approving full path as sometimes Windows has issues with the `notepad` command.

PR #790, opened 2021-04-02
Correct and add alt text

karenword:
One missed addition and a fix from the Core Team acc-athon


munkm (approved):
This looks great! Thank you for the clarifiying alt text.

PR #793, opened 2021-04-07
Update 03-create.md

paulcraig:
I propose changing the home branch from master to main in keeping with current GitHub practice. I am submitting this as part of my teaching certification for Software Carpentries.

PR #798, opened 2021-04-07
Update 08-collab.md

paulcraig:
I am proposing changing 'master' to 'main' 12 times in this document in keeping with current GitHub practice for the root branch for a git repository. This is part of my certification for becoming a Software Carpentries instructor.


paulcraig:
Merged these comments with #793

Issue #761, opened 2020-10-06
Best practices to switch from master to main


samteplitzky:
As Github is switching its terminology from master to main (https://github.com/github/renaming) over the next several months, I'm curious what the Carpentries approach will be. In this lesson, there are several episodes (7, 8, 9 on quick glance) that would require quick changes from "master" to "main." Is this a current consideration?

zingale:
I just taught the git lessons today, and indeed, when you create the github repo, it now defaults to `main`, and the instructions for pushing your changes also refer to `main`.  There was no issue (at the moment) just using `master`, as it just seemed to work.  But no one was using the latest version of git, which also apparently defaults to `main`.

sstevens2:
For the latest version of git, from what I've seen you still need to configure your default branch to be `main` with something like `git config --global init.defaultBranch main`.

I taught git last week and honestly I installed git with xcode on this computer and didn't want to try to figure out how to update before the lesson.  So instead, I did `git checkout -b main` as a setup step after initializing the repo (and explained why) and since there were no commits on `master` it doesn't exist anymore. The only person in the class this didn't work for was someone who had git 1.XX version instead.  So they had to commit first and then I had them switch branches and delete the `master` branch later.

```
cd planets
git init
git checkout -b main
```
Then your branch is named `main` and the `master` branch doesn't exist.  

This also was a bit more relevant for me because I teach a [custom version that teaches branches and PRs](https://uw-madison-datascience.github.io/git-novice-custom/) so we need to talk about branches anyway.  I still presented it as a setup step and explained that master is a historical reference to slavery and as a community we are moving to the default branch being called `main` instead.

chendaniely:
If you do forget to change your branch names and `master` is already been commited to. you can use `git branch -m main` which will rename the current branch you are on. or more explititly renaming master to main via `git branch -m master main`.

this should reduce the git-mastrics of creating and deleting branches in separate steps (which tbh is what I normally do)

reference: https://stackoverflow.com/questions/6591213/how-do-i-rename-a-local-git-branch

zkamvar:
FWIW, the new version of the git installer for windows now gives the users an option to explictly switch their default branch to 'main': https://github.com/carpentries/workshop-template/pull/711

sstevens2:
Should we add a step to change the branch to `main` after `git init` for the folks who are on mac/linux systems?  I can draft a PR tomorrow for review if so.

ha0ye:
[...]

PR #804, opened 2021-05-17
Update to main branch after init

sstevens2:
Based on the discussion in issue #761.  Created a PR that switches to the main branch after initializing.  Text edits welcome!

Also added in setup episode suggestion from @srerickson.  @srerickson let me know if you'd prefer I revert that commit so you can get credit in the history.

This does not address all the subsequent times that the branch name appears in the status command.  Looks like there are already a few PRs on that topic #788 and #793

### DIFF
--- a/_episodes/02-setup.md
+++ b/_episodes/02-setup.md
@@ -146,7 +146,8 @@ same commands to choose another editor or update your email address.
 > when you create a new repository with `git init` (as explained in the next Episode). This term evokes 
 > the racist practice of human slavery and the software development community has moved to adopt 
 > more inclusive language. Notably, new repositories created through GitHub use `main` as the default
-> branch name.
+> branch name.  
+> Read more about how to implement this change in [GitHub's renaming repository](https://github.com/github/renaming).
 >
 > With Git versions 2.28 and higher, you can set a different name for the initial branch. To set main as 
 > the default branch name do:
--- a/_episodes/02-setup.md
+++ b/_episodes/02-setup.md
@@ -116,6 +116,14 @@ $ git config --global init.defaultBranch main
 
 > ## Default Git branch naming
 >
+> Source file changes are associated with a "branch." 
+> For new learners in this lesson, it's enough to know that branches exist, and this lesson uses one branch.  
+> By default, Git will create a branch called `master` 
+> when you create a new repository with `git init` (as explained in the next Episode). This term evokes 
+> the racist practice of human slavery and the 
+> [software development community](https://github.com/github/renaming)  has moved to adopt 
+> more inclusive language. 
+> 
 > In 2020, most Git code hosting services transitioned to using `main` as the default 
 > branch. As an example, any new repository that is opened in GitHub and GitLab default 
 > to `main`.  However, Git has not yet made the same change.  As a result, local repositories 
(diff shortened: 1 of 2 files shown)


srerickson:
thanks for offering to revert, but not necessary! thanks for putting this in a PR!

kekoziar (changes requested):
Can you update your episode 2 file to the most current file?  I merged a different PR for this because it kept the context of "Dracula decides to [...]."  However, I'd like to include some of your content.  Specifically, 

> Source file changes are associated with a "branch." By default, Git will create a branch called `master` when you create a new repository with `git init` (as explained in the next Episode). This term evokes the racist practice of human slavery and the software development community has moved to adopt more inclusive language. 

and, perhaps insert a link to https://github.com/github/renaming?

sstevens2:
Hi @ kekoziar. I think I made the changes you suggested.  Additional feedback welcome!

_episodes/03-create.md
```diff
@@ -85,6 +85,21 @@ including all files and sub-directories located within the project's directory.
 If we ever delete the `.git` subdirectory,
 we will lose the project's history.
 
+Next, we will change the default branch to be called `main`.
+This might be the default branch depending on your settings and version
```
nsoranzo, line 89:
```suggestion
This might already be the default branch name depending on your settings and version
```

_episodes/03-create.md
```diff
@@ -85,6 +85,21 @@ including all files and sub-directories located within the project's directory.
...
+of git.
```
nsoranzo, line 90:
```suggestion
of Git.
```

_episodes/02-setup.md
```diff
@@ -140,6 +140,23 @@ $ git config --list
 You can change your configuration as many times as you want: use the
 same commands to choose another editor or update your email address.
 
+> ## Default Branch Name
```
nsoranzo, line 143: I think this part needs to be merged with what has already been added as lines 108-128.

kekoziar:
Hi, @sstevens2.  Thanks for updating your PR with the current repository.  The intention is to put the first few sentences in the 
"Default Git branch naming" callout that's you pulled in with the update.  I think the link can pair with the phrase "software development community".

Also, as a note, callouts require `{: .callout}` at the end of the section.  You can see the example in the callout I referenced.  

Does this all make sense?

kekoziar:
verified changes.

michaelberks:
I'm not sure if this is the best place to leave this comment (there are several open issues re the switch from 'master' to 'main', but I wonder if having set main as the default branch in episode 2: setting up git, we should remove the checkout to main in episode 3?

why:
1. Trying to switch to a branch we're already on isn't normal git practice, and would usually generate an error. The only reason it doesn't throw an error here is because there are no commits yet on main
2. It means the user doesn't know if their global setting has actually worked (because we'll now be on main anyway). Whereas if we miss out the checkout, then when we instruct the user to call 'git status', we can ask them to verify it shows 'main', and if doesn't get them to check their config settings
3. It means we don't have to introduce branch checking out at all, which given we're not covering branches elsewhere in the lesson eases cognitive load

If anyone else agrees, I can put in a PR with suggesting changes (ie dropping the 'git checkout -b main', and adding some text on getting the user to checkout the output from their first 'git status' shows the branch as main)

PR #806, opened 2021-05-20
Add notepad.exe as git text editor

sshumway:
For windows users who prefer to avoid installing a third-party text editor, notepad.exe is a viable option for git's default editor. Added a row to the editors table for setting notepad.exe as the `core.editor`.


kekoziar:
duplicates https://github.com/swcarpentry/git-novice/pull/789

PR #808, opened 2021-05-22
ENH: Show default branch name command

### LINKED ISSUE #805: Transition to `main` as default branch name?
There as been a heated debated in the past few months about the convenience of transitioning the default branch name from `master` to `main`.

Although `git` itself has not made the transition yet, strides are being made towards that purpose:
https://lore.kernel.org/git/pull.762.git.1605221038.gitgitgadget@gmail.com/

`GitHub`, `GitLab` and other major code hosting services already use `main` as the default branch name when starting a new repository.

So the evidence is that the transition has been happening. Maybe the lesson should be updated to match the new naming, or at least the transition could be mentioned as a callout so that users are warned about it in case their local `git` version has still not transitioned.

Loosely related to #716.

Thanks for this wonderful lesson

jhlegarreta:
Show how to set a default branch name when configuring `git`.

Add a note in the instructor notes to change the default branch name for the
lesson if learners have a pre-2.28 Git version.

Transition from `master` to `main` as the default branch name.

Fixes #805.

### DIFF
--- a/_episodes/02-setup.md
+++ b/_episodes/02-setup.md
@@ -105,8 +105,9 @@ your changes, press <kbd>Esc</kbd> then type `:q!` and hit <kbd>Return</kbd>.
 > If you want to save your changes and quit, press <kbd>Esc</kbd> then type `:wq` and hit <kbd>Return</kbd>.
 {: .callout}
 
-Recent versions of Git (since 2.28) allow to configure the name of the branch created when you
-initialize any new repository. Dracula decides to use that feature to set it to `main`:
+Git (2.28+) allows configuration of the name of the branch created when you
+initialize any new repository.  Dracula decides to use that feature to set it to `main` so 
+it matches the cloud service he will eventually use. 
 
 ~~~
 $ git config --global init.defaultBranch main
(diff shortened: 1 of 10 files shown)


jhlegarreta:
Overlaps with #804 and #793. Should have read the open PRs before opening the issue and making the PR. Sorry.

Compared to #804, this change:
- Explicitly makes all learners to set their default branch name to `main` to avoid different setups, as suggested in https://github.com/swcarpentry/git-novice/issues/805#issuecomment-846442638. As a consequence, updates this passage https://github.com/swcarpentry/git-novice/pull/808/files#diff-46347e9b0264525f307194a749587e5b2e9554e4b3b37c281135e7af40ed8ccaR129.
- Provides the context of the change in a `callout`.

@sstevens2 @srerickson maybe this is worthwhile considering. Otherwise, please go ahead and close this PR. Thanks for your time, and again, sorry for the duplicate effort.

sstevens2:
I'll leave it up to the maintainers.  The solution in #804 will work for older versions of git where setting the default branch is only an option on newer versions of git.

kekoziar:
@jhlegarreta and @sstevens2, indeed your two PRs complement each other and I plan on merging them.  I proposed some text changes to this one which reflect https://github.com/swcarpentry/git-novice/pull/804.  

I also proposed deletion of the reference to https://lore.kernel.org/git/xmqq5za8hpir.fsf@gitster.c.googlers.com/, since that adds a significant amount of cognitive load.  There might be a case that it can be added to the instructor notes, but I'm inclined to not include it since the git patch notes aren't documented in the lesson.

kekoziar (approved):
Thanks for submitting these changes and especially including the global command to rename the default branch.

jhlegarreta:
@kekoziar :+1:. I can propose new PRs to address the remaining items in #812. Let me know what works best for you.

Issue #812, opened 2021-06-28
Checklist: Replace `master` with `main`


kekoziar:
There are several, sometimes overlapping, issues and PRs regarding replacing `master` with `main` to align with GitHub's terminology.  GitHub made this change last year, but git has yet to follow suit.  

I think it’s a good idea to make the change in line with GitHub’s create new repository CLI commands.  I think the changes can be implemented without having to explain a lot about branches and creating cognitive overload by saying that as a beginning learner in git, 
1.  it’s enough that learners know branches exist, but we’re only using one branch (`main`) in the lesson; 
2.  branches in an existing repository that are in use by other collaborators should not be renamed w/o discussion with the group (per https://git-scm.com/book/en/v2/Git-Branching-Branch-Management; perhaps this content can go in instructors notes); and,
3.  branch management is an advanced topic.

As an inventory of changes which have been suggested and must take place to reflect the replacement

- [x]  Add command line instruction to update `master` with `main`. (`git checkout -b main`)
- [x]  Create call-out that includes some background why this change happened.  
- [x]  Replace references of `master` with `main`.
- [x]  Update screenshots which reference `master`
- [x]  References in extras section updated

current issues:
https://github.com/swcarpentry/git-novice/issues/761, https://github.com/swcarpentry/git-novice/issues/775, https://github.com/swcarpentry/git-novice/issues/776, https://github.com/swcarpentry/git-novice/issues/805

tobyhodges:
Thank you for opening up this issue @kekoziar. As you mentioned, the transition to `main` has already taken place on GitHub and in several other places (e.g. GitLab, Git for Windows), and (as many contributors have noted in the other issues and pull requests around this topic) this is a shift that reflects [the values of The Carpentries](https://carpentries.org/values/). There are multiple Pull Requests open (#793 #788 #804 and #808) to update the language in the lesson. The Carpentries Curriculum Team would be delighted if you and/or the other @swcarpentry/git-novice-maintainers can take some time to review them, merge, request changes, close duplicates, as you consider best for the flow of the lesson. If there is anything that the Core Team can do to support you in this, please let me know here.

I will try my best to outline what has been tackled in each of those open Pull Requests:

- #788 & #793: changing mentions of `master` to `main`
- #804: update the default branch name in episode 3; configure the default (depending on Git version) and explain the change on the lesson Setup page
[...]

PR #820, opened 2021-07-20
Updated the cartoon in Episode 5: branch name to main; better representation of version control

yangmr04:
This is a contribution to #812 as the branch name is still "master" in the cartoon.

Fixed some graphics by putting borders around the files.

I also modified the content a bit, I think it represents git better, at least now everything goes through the version control guy instead of bypassing them.


kekoziar:
Thanks, @yangmr04.  I really appreciate your improvement to the lesson.  I made some minor adjustments, but it looks good!

PR #825, opened 2021-07-25
Adjust nomenclature for "episode", "lesson", and "section" (fixes #777)

### LINKED ISSUE #777: "Lesson" is used in place of "episode"
I believe the standard Carpentries nomenclature is that a _lesson_ is a collection of _episodes_—so this repository contains the git-novice _lesson_, which comprises _episodes_ such as "Setting up Git", "Creating a Repository", "Tracking Changes", etc.

This terminology is mixed up in the lesson material, however. Episodes 7 and 8 refer to episodes, but elsewhere in the lesson the episodes are referred to as "lessons" - specifically:

* In episode 2, "in a later lesson", "For these lessons"
* In episode 5, 'As we saw in the previous lesson"
* In episode 7, "If you remember back to the earlier lesson", "In this lesson" (and also uses "section"—"in the final section", "In this section"—to refer to the episode)
* In episode 10: "You'll learn how to do this in lesson 12: Citation"

I think in the minimal case that the nomenclature should be consistent across all episodes of the lesson, rather than the mixing that is currently done. Making it consistent with the Carpentries usual nomenclature seems like the best option?

edbennett:
As discussed in issue #777, some of the existing wording was inconsistent with the Carpentries' nomenclature, which could cause confusion for some. This PR adjusts the wording to be consistent, and in cases where the meaning was already or became ambiguous as a result of this change, adjusts the surrounding wording to clarify what I believe the intended meaning was. More detail on each change is in the commit log

### DIFF
--- a/_episodes/02-setup.md
+++ b/_episodes/02-setup.md
@@ -35,9 +35,9 @@ which means that any changes pushed to
 [BitBucket](https://bitbucket.org/),
 [GitLab](https://gitlab.com/) or
 another Git host server
-in a later lesson will include this information.
+after this lesson will include this information.
 
-For these lessons, we will be interacting with [GitHub](https://github.com/) and so the email address used should be the same as the one used when setting up your GitHub account. If you are concerned about privacy, please review [GitHub's instructions for keeping your email address private][git-privacy]. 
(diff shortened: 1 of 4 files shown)


kekoziar:
Thanks, @edbennett, for the PR!

kekoziar (approved):

PR #828, opened 2021-08-05
Fix typo in 'right-hand'

alisonrclarke:
Spotted a typo whilst looking through the lesson.


kekoziar (approved):
@alisonrclarke Thanks for catching the typo!

PR #829, opened 2021-08-09
replace self-describing edits with representative ones [fix #727]

### LINKED ISSUE #727: Should text in "conflicts" examples be representative rather than self-describing
Up until the episode on [conflicts](http://swcarpentry.github.io/git-novice/09-conflict/index.html), all examples are representative of how git might be used in a real workflow, with Dracula making notes vaguely related to the hypothetical scenario being worked on. However, in the Conflicts episode, this abruptly switches to self-describing examples, with text like "This line added to Wolfman's copy"—this is not motivated, and is likely to cause additional cognitive load on learners (and also makes it harder for instructors to teach it). When the merge conflict is resolved, this makes it less clear to the learner what the resolution means.

I feel it would be better to have two relatively motivated "in-universe" changes made by Wolfman and Dracula that are inconsistent with each other, then these are resolved by finding a middle ground.

Even better would be if the changes were each two-line changes, and only one line was inconsistent, so that we could show that some changes could be kept—this would make it clearer why both versions are shown in the editor. Otherwise we may be giving the mistaken impression that you have to delete all conflicting lines and start from scratch.

(I'm happy to draft the changes to the lesson if we agree that this is something worth doing.)

edbennett:
As discussed in #727, this PR changes the text used when creating and resolving a merge conflict, so that it is clearer to the learner how a conflict may occur in their everyday work, and how an approach to resolving it may look.

In the existing material, the change from "in-universe" edits to self-describing ones may be jarring, and makes the workflow seem more forced. Changing it to match the material in the previous sections will hopefully smooth this over.

It also illustrates the thought process behind resolving a conflict, which isn't just to delete whatever is in the conflict and write something new (which the current example may seem to imply), but rather considering the contributions made by both conflicting commits and ensuring that the resulting merged commit includes/reflects both.

### DIFF
--- a/_episodes/09-conflict.md
+++ b/_episodes/09-conflict.md
@@ -47,7 +47,8 @@ $ cat mars.txt
 Cold and dry, but everything is my favorite color
 The two moons may be a problem for Wolfman
 But the Mummy will appreciate the lack of humidity
-This line added to Wolfman's copy
+An underground base with few windows would be best, to minimise Wolfman's exposure to the moons
+A windbreak could protect a base from dust storms
 ~~~
 {: .output}
 
@@ -55,13 +56,13 @@ and then push the change to GitHub:
 
 ~~~
 $ git add mars.txt
-$ git commit -m "Add a line in our home copy"
+$ git commit -m "Add discussion of base type and weather protections"
 ~~~
 {: .language-bash}
 
 ~~~
-[main 5ae9631] Add a line in our home copy
- 1 file changed, 1 insertion(+)
+[main 5ae9631] Add discussion of base type and weather protections
+ 1 file changed, 2 insertions(+)
 ~~~
 {: .output}
 
@@ -97,7 +98,8 @@ $ cat mars.txt
 Cold and dry, but everything is my favorite color
 The two moons may be a problem for Wolfman
 But the Mummy will appreciate the lack of humidity
-We added a different line in the other copy
+A tall, glass base would be best, to give Vlad the best view of the red surroundings
+A helipad could be useful for supplies arriving
 ~~~
 {: .output}
 
@@ -105,13 +107,13 @@ We can commit the change locally:
 
 ~~~
 $ git add mars.txt
-$ git commit -m "Add a line in my copy"
+$ git commit -m "Add points around base layout"
 ~~~
 {: .language-bash}
 
 ~~~
-[main 07ebc69] Add a line in my copy
- 1 file changed, 1 insertion(+)
+[main 07ebc69] Add points around base layout
+ 1 file changed, 2 insertions(+)
 ~~~
 {: .output}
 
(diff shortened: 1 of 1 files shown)


_episodes/09-conflict.md
```diff
@@ -47,21 +47,22 @@ $ cat mars.txt
 Cold and dry, but everything is my favorite color
 The two moons may be a problem for Wolfman
 But the Mummy will appreciate the lack of humidity
-This line added to Wolfman's copy
+An underground base with few windows would be best, to minimise Wolfman's exposure to the moons
```
kekoziar, line 50: These lines are too long.  They should be shorter, and add only one line at a time.  This allows the learner to focus on how Git operates instead of typing out all the extraneous details of the story. 

Maybe

> An underground base to minimise Wolfman's exposure to the moons

_episodes/09-conflict.md
```diff
@@ -47,21 +47,22 @@ $ cat mars.txt
...
+A windbreak could protect a base from dust storms
 ~~~
 {: .output}
 
 and then push the change to GitHub:
 
 ~~~
 $ git add mars.txt
-$ git commit -m "Add a line in our home copy"
+$ git commit -m "Add discussion of base type and weather protections"
```
kekoziar, line 59: I'd prefer something that's in the story but is easy for a novice learner to see that this line was added from Wolfman's copy.

_episodes/09-conflict.md
```diff
@@ -97,21 +98,22 @@ $ cat mars.txt
...
+A helipad could be useful for supplies arriving
 ~~~
 {: .output}
 
 We can commit the change locally:
 
 ~~~
 $ git add mars.txt
-$ git commit -m "Add a line in my copy"
+$ git commit -m "Add points around base layout"
```
kekoziar, line 110: Again, something in the story but is easy for a novice learner to see that this line was added from Vlad's copy. Maybe

> Vlad's suggestions on base layout

kekoziar (changes requested):
@edbennett, thanks for the PR.  I made some suggestions which boil down to 

- shorter lines added to the `mars.txt` file - I think no more than nine words (otherwise it takes to long to type everything); and, 
- making it readily apparent who made the changes in the commit messages.

_episodes/09-conflict.md
edbennett: I agree these lines are too long; I was trying to work out how to make it clear who wrote them while also making a point relevant to the story. Thanks for the suggestion of a shorter alternative.

I disagree that there should only be one line in each change. My concern is that this will give the impression that the way to deal with a merge conflict is "delete every conflicting line that Git has highlighted and start again". The vast majority of merge conflicts will involve leaving some lines of the original present; I think having the second line present makes it much clearer what the thought process behind the conflict resolution is.

_episodes/09-conflict.md
edbennett: This would subvert the intent of this pull request. The point is to demonstrate good git commit practice as much as possible, since excluding exercises, in the full lesson a learner will only see an instructor make 9 commits. If two of those include data around who authored the commit (which shouldn't be in a commit message, because that data is recorded separately by Git—as we discuss in episode 4), then there is a risk that novices will pattern their commit messages on this style and include redundant information in their commit messages. This risk is elevated because these are the last two commits that a learner will see an instructor make in this lesson.

I"m not sure that having difficulty following who is playing who, or whose commits are whose, in this section is specific to being a novice Git user. Would it be better to add extra expository text before the commit (line 104/106 in the diff above) to reaffirm that this is specific to Vlad/the repository owner's copy.

kekoziar:
I'm going to close this PR in preparation of the new workbench lesson infrastructure, but with a comment that this may be a moot point, as the story used for the lesson may be changing.

PR #830, opened 2021-08-12
Update 07-github.md

annajiat:
1) Updated challenge title to match with the challenge itself.
2) In case of the learner facing the same problem due to another reason e.g. adding another file from the web interface, the solution can now be located using the web browser's find feature CTRL+F, instead of being hidden away in the collapsed solution.

Proposed:
> ## Fixing "refusing to merge unrelated histories" in relation to GitHub License and README files

Current:
> ## GitHub License and README files

### DIFF
--- a/_episodes/07-github.md
+++ b/_episodes/07-github.md
@@ -443,7 +443,7 @@ GitHub, though, this command would download them to our local repository.
 > {: .solution}
 {: .challenge}
 
-> ## GitHub License and README files
+> ## Fixing "refusing to merge unrelated histories" in relation to GitHub License and README files
 >
 > In this episode we learned about creating a remote repository on GitHub, but when you initialized 
 > your GitHub repo, you didn't add a README.md or a license file. If you had, what do you think 


_episodes/07-github.md
```diff
@@ -443,7 +443,7 @@ GitHub, though, this command would download them to our local repository.
 > {: .solution}
 {: .challenge}
 
-> ## GitHub License and README files
+> ## Fixing "refusing to merge unrelated histories" in relation to GitHub License and README files
```
kekoziar, line 446: The episode already refers to the title of this challenge

> See the “GitHub License and README files” exercise below for a full explanation of why the repository needs to be empty.

I don't see how this is an improvement. If a novice learner is walking through this, they will not encounter the situation you described.

annajiat: Thank you for reviewing. 

There could be various reasons for encountering it.  During the workshop the note may be missed perhaps due to being in between two large screenshots or more importantly while the learners are reviewing the materials after the workshop, they might not be able to locate "refusing to merge unrelated histories" by searching for it in the lesson. The web browser's search (CTRL+F or Cmd+F) does not work because the solution of exercise needs to be expanded first.

Alternatively, we could add the following line to the note so that CTRL+F works and also highlight the note:
to avoid  the error "refusing to merge unrelated histories"

So, it will look like:
**Note**: Since this repository will be connected to a local repository, it needs to be empty. Leave “Initialize this repository with a README” unchecked, and keep “None” as options for both “Add .gitignore” and “Add a license” **to avoid  the error "refusing to merge unrelated histories"**. See the “[GitHub License and README files](https://swcarpentry.github.io/git-novice/07-github/index.html#:~:text=tried%20to%20link%20your%20local%20and%20remote%20repositories)” exercise below for a full explanation of why the repository needs to be empty.

Should we go ahead with this alternative?

Additionally, we may also choose to make it a yellow **callout** to further highlight it and for the ease of referring to it during the workshop via the link to the **callout**.

scottgruber:
Hi, I'm a new maintainer to this repository and reviewing pull requests and issues. If I follow this thread correctly, the maintainer makes a point that the novice learners wouldn't encounter the situation described. I'll close this PR, however @annajiat if you have any further comments please reopen this PR.

PR #833, opened 2021-08-13
Update 02-setup.md

### LINKED ISSUE #809: Mention Enter in addition to Return in the Git lesson
There are a few mentions of pressing Return in the lesson in https://swcarpentry.github.io/git-novice/02-setup/index.html:

* As with other keys, when you hit Return on your keyboard...
* If you haven’t used Vim before and wish to exit a session without saving your changes, press Esc then type :q! and hit Return

In many keyboard layouts, Return is labelled as Enter or even a symbol. I'd suggest labelling the key as Return / Enter / ↵

annajiat:
Closes #809 
Mentioned **Enter** key and its symbol in addition to the **Return** key

### DIFF
--- a/_episodes/02-setup.md
+++ b/_episodes/02-setup.md
@@ -47,7 +47,7 @@ For this lesson, we will be interacting with [GitHub](https://github.com/) and s
 
 > ## Line Endings
 >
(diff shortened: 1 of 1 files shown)


kekoziar (approved):
looks good.  Thanks for the contribution.

PR #836, opened 2021-08-13
Update 08-collab.md

### LINKED ISSUE #724: Episode 8 - No notifications in Github after inviting collaborator
I recently co-instructed on the Git lesson. We asked participants to join into pairs, and pick roles of collaborator and owner. After the owner had invited the collaborator, we did not find an easily accessible place in Github to accept this invitation. There was no notification at the link given in the lesson, https://github.com/notifications (which redirects to https://github.com/notifications/beta).

When the collaborator received an email, it worked to accept from there. However, some had a problem receiving the email.

We found that the collaborator could go to `https://github.com/USER/planets/invitations` , where `USER` is the name of the owner. I'm not completely satisfied with this solution, because it requires them to input a long URL manually.

Do others have the same experience, or does it usually work better?

annajiat:
Closes #724

Added note for checking email in addition to checking notification.

### DIFF
--- a/_episodes/08-collab.md
+++ b/_episodes/08-collab.md
@@ -32,8 +32,8 @@ then enter your partner's username.
 ![Adding Collaborators on GitHub](../fig/github-add-collaborators.png)
 
 To accept access to the Owner's repo, the Collaborator
-needs to go to [https://github.com/notifications](https://github.com/notifications).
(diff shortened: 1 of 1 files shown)


kekoziar (approved):
Looks good.  Thanks for the contribution!

PR #837, opened 2021-08-13
Update 04-changes.md

### LINKED ISSUE #695: Incomplete output from `cat` in Episode 4, "Committing Multiple Files"
The first step of the "Committing Multiple Files" exercise in [Episode 4](http://swcarpentry.github.io/git-novice/04-changes/index.html) is to add a new line of text to `mars.txt`. The command `cat mars.txt` is then used to examine the contents of the file before proceeding.

The solution shows that the output of the `cat` command should be
```
Maybe I should start with a base on Venus.
```
but it actually shows all the text in `mars.txt`, which at this point is
```
Cold and dry, but everything is my favorite color
The two moons may be a problem for Wolfman
But the Mummy will appreciate the lack of humidity
Maybe I should start a base on Venus.
```

When I went through the exercise I wasn't sure whether the text had intentionally been cut for brevity, was unintentionally incomplete or whether I should have replaced the earlier lines in the file. (I'd guess not the last option as the solution's commit log says "2 insertions" only.) If someone could clarify, I'd be happy to open a PR to fix 😄

annajiat:
Closes #695

### DIFF
--- a/_episodes/04-changes.md
+++ b/_episodes/04-changes.md
@@ -682,6 +682,7 @@ repository (`git commit`):
 > >
 > > The output below from `cat mars.txt` reflects only content added during 
 > > this exercise. Your output may vary.
+> > 
 > > First we make our changes to the `mars.txt` and `venus.txt` files:
 > > ~~~
 > > $ nano mars.txt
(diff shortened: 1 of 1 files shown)


kekoziar (approved):
Changes look good.  Thank you for contributing!

Issue #781, opened 2021-02-16
Improving the setup page


Madchild95:
Hello community

I have made a proof reading of the [Git novice lesson](http://swcarpentry.github.io/git-novice/) and thought I would make a comment on the [setup page](http://swcarpentry.github.io/git-novice/setup.html). I think the issue has already been discussed in here: #735, but I wanted to make a suggestion for how to make it a bit better, while still keeping the link for the [general setup page](https://carpentries.github.io/workshop-template/#git  ) (as I gathered that it was important to keep the setup page for the lesson and the instruction page for the setup separate).

As of now the current setup page looks like this:
<img width="1161" alt="Screenshot 2021-02-16 at 11 34 31" src="https://user-images.githubusercontent.com/54779393/108051575-596b6280-704b-11eb-9ef9-3bb7e87535f3.png">

**Issue:** Personally, the first thing I would gather from this page is that setup for this lesson is just to have the Desktop as working directory in the terminal. Whereas the setup that really matters, namely to install Git and GitHub is sort of in the background and just mentioned in one sentence with a hyperlink.

**Solution:** My suggestion is to make a heading that says "_Installing Git and creating GitHub account_" much like in other lessons such as the [Plotting and Programming with Python setup page](http://swcarpentry.github.io/python-novice-gapminder/setup.html):
<img width="1163" alt="Screenshot 2021-02-16 at 11 45 23" src="https://user-images.githubusercontent.com/54779393/108052608-b0be0280-704c-11eb-8434-71550e9b8806.png">

An explanation in this section could include an introduction text like the one on the general setup page: 
"Git is a version control system that lets you track who made changes to what when and has options for easily updating a shared or public version of your code on github.com. You will need a supported web browser.

You will need an account at github.com for parts of the Git lesson. Basic GitHub accounts are free. We encourage you to create a GitHub account if you don't have one already. Please consider what personal information you'd like to reveal. For example, you may want to review these [instructions for keeping your email address private](https://docs.github.com/en/github/setting-up-and-managing-your-github-user-account/setting-your-commit-email-address) provided at GitHub."

And then additional information about how to navigate in the general setup page as a warning, something like:
[...]

PR #842, opened 2021-10-04
Add numbered headers to avoid skipping steps

marco-foscato:
The numbered headers are meant to improve the visibility of the steps in the procedure needed to successfully configure and use the remote repo. This change came up as a suggestion in the debriefing following a lesson during which the crucial step where the public key is copied to GitHub went unnoticed and was accidentally skipped.


marco-foscato:
This could be a fix for #781

kekoziar (approved):
I fixed a couple of typos, otherwise the changes look good.  Thank you, they definitely improve readability for this episode!

Issue #847, opened 2021-10-31
Git Novice Episode 2 - give more detail about text editors


k-tatgenhorst:
The lessons are well taught, but I feel one area lacking detail is about the choice of editor. 

How to set the editor is of course mentioned, but then additional information is listed for VIM which leads a novice to think this is an implied preference choice for the course. I feel that text above the selection mentioning the fact that many linux/unix users have editor preferences and accumulated skills which foster their efficiency should be mentioned. Additionally, the student might be given the suggestion to search outside of class for details on incorporating their choice of editor into their git workflow.

I do not believe this would add to the required length of time for the lesson, only that it would reinforce that editor selection is a personal choice.

Thank you,
Karl Tatgenhorst

michaelberks:
I actually recommend taking the opposite approach, and removing all references to setting Git's default editor from episode 2.

My reasoning:
1. There's no example in the remainder of the lesson where Git's editor is used
2. The only time in normal workflows the editor gets used is when asked to comment on a merge conflict. This is a fairly rare occurrence, particularly in the likely use cases of git novices. Note this is taking account that:
    - commit messages are taught using the "-m" option and setting a one line message (best practice for small, regular, local commits) and note post-edited once set
    - the only time longer commit messages are recommended are *eg* squashing commits into pull-requests, which are beyond the scope of this tutorial and more normally done directly in GitHub's (or GitLab etc) web-interface
3. Because it's used rarely, setting the default as a config doesn't seem particularly important. *eg* I've used Git for years, including in some fairly complex GitFlow settings, and had never thought to set this config until I taught this lesson.
4. Git's default behaviour if the config is unset seems preferable to recommending **vim** anyway
    - If the user has set a default editor for their shell Git will use that (presumably what the user in this case would prefer)
    - If no shell default is found, Git will fall back to using **vi**, which is close enough to **vim** for the editing tasks you're likely to do in Git
5. One of the principals of carpentries lessons is "get to the most useful stuff as soon as possible". I'd argue in this case this means spending as little time as necessary in setup steps of episode 2.
6. The whole lesson is really hard to teach in the given timescale, therefore any material not required in the context of the lesson or in likely novice use cases following the lesson is helpful to remove (especially when other things get added, *eg* the change of default branch from master to main).

I'm sure others may disagree, but as I need to make a lesson PR as part of my carpentries instructor checkout process, I have made a PR with the editor section removed. I understand totally if this PR gets rejected.

[...]

PR #852, opened 2021-11-24
Remove setting default editor from episode 2

michaelberks:
This change addresses #847 and recommends removing the section on setting Git's default editor from episode 2 of the lesson, with an additional small change in the initial bullet point list at the start of the lesson, replacing the reference to setting the editor with "choosing how we deal with line endings in text files". My reasoning for this change is give in my comment on the issue [here](https://github.com/swcarpentry/git-novice/issues/847#issuecomment-977867474).

I have checked the lesson still compiles and formats correctly by rebuilding and running on a local server.

This PR has been made as a contribution towards instructor training.

### DIFF
--- a/_episodes/02-setup.md
+++ b/_episodes/02-setup.md
@@ -16,7 +16,7 @@ we need to configure a few things. Below are a few examples
 of configurations we will set as we get started with Git:
 
 *   our name and email address,
-*   what our preferred text editor is,
+*   choosing how we deal with line endings in text files,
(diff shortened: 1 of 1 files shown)


kekoziar (commented):
Text editor information is included to counter anyone executing `git commit` with no `-m` and message. This usually happens at least once during a workshop.  The most recent time I had to walk someone out of vim because they executed `git commit` was last week.  Removing this information doesn't add, but only makes the lesson more difficult.

kekoziar:
Closing.

PR #853, opened 2021-11-30
Updates git clone to use SSH rather than HTTPS

pgmccann:
For consistency with the previous episode, and avoiding authentication errors.


kekoziar (approved):
verified changes.  thanks for the update!

PR #856, opened 2021-12-06
some wording change suggestions

slayoo:
- avoiding the word "move" when changing directory, not to confuse with moving data
- avoiding "all" wording for files and subdirectories within a git-tracked folder, not to mislead a reader that everything is automatically tracked


scottgruber (approved):
Thank you for helping improve the wording.

PR #859, opened 2022-01-12
Replace "ridiculous" with "unnecessary" and added links to RCS, CVS, subversion

dcassol:
- Replaced "ridiculous" with "unnecessary". 
- Added links to find more information about `RCS`, `CVS`, and `subversion`


kekoziar (approved):
@dcassol, thanks for the contribution.  I like the change from _ridiculous_ to _unnecessary_  because _unnecessary_ is less likely to be used in an emotional context.  And, links to tools and systems that are referenced but not explained are always appreciated. 

PR approved.

PR #869, opened 2022-03-30
Improve figures page for sw-Git lesson

goi42:
Add the neglected Universal Missions image and sort images according to first appearance in the lesson


kekoziar:
`_includes/all_figures.html` is an automatically generated file.  This should be resolved with the new infrastructure.

PR #872, opened 2022-04-02
Add clearer Git install instructions

wdconinc:
As discussed in #781 the instructions are dominated by the instructions for preparing a directory, to the point that the instructions for installing the Git command are easily overlooked. This PR gives additional visual weight to the Git command installation, adds headers, while maintaining reference to the central instructions.

image

Potential for improvement:
- The three OS-specific links are identical. Deep-linking to the OS-specific git-windows, git-macos, and git-linux anchors only works when those are opened by default, so really only for git-windows. Furthermore, those deep links jump to the content of the box with OS-specific instructions, so the Git header and preamble in that section are not visible anymore. Ideally, #git-macos would open the menu on MacOS instructions but still link to the #git anchor. Not sure that's possible.

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -3,15 +3,15 @@ layout: page
 title: Setup
 ---
 
-## Install Git
+## Installing Git
 Since several Carpentries lessons rely on Git, please see
 [this section of the workshop template][workshop-setup] for
 instructions on installing Git for various operating systems.
-- [Git installation on Windows][workshop-setup-windows]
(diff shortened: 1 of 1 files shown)


emcaulay (approved):
Having just been a helper on this workshop, I can vouch for the need for giving a little more information on this git setup page. I was really confused even as a helper and just needed a little more information on this page. This edit is a great help!

kekoziar (approved):
changes look good.  Thanks!

PR #880, opened 2022-06-07
Changed the Line Endings git config option so it works on Windows

catherinef37:
Our group noticed at a recent workshop that in Git bash for Windows, setting `git config --global core.autocrlf false` would prevent the error messages about line endings (i.e. warning: LF will be replaced by CRLF in hello.txt. The file will have its original line endings in your working directory). Currently the instructions have this option set to `true`, which does not do what we want.

### DIFF
--- a/_episodes/02-setup.md
+++ b/_episodes/02-setup.md
@@ -70,7 +70,7 @@ For this lesson, we will be interacting with [GitHub](https://github.com/) and s
 > And on Windows:
 >
 > ~~~
-> $ git config --global core.autocrlf true
+> $ git config --global core.autocrlf false
 > ~~~
 > {: .language-bash}
 > 


scottgruber:
Hi Catherine,

Thank you for submitting a PR. Do you know what effect setting the option to false with have on macOS? If you don't know please comment and I will take a look. 

Thank you for your help. 

Scott

catherinef37:
Hi Scott,

I unfortunately don't know, since most of our group used Windows and Linux, and I don't have access to a device with macOS. That's a great thing to check, thanks so much for taking a look at this!

Catherine

kekoziar (approved):
verified change.  approved.

dpshelio:
@kekoziar - I believe this change should be reverted! This will increase the difficulty of collaboration between different systems. The error message that @catherinef37 refers is not an error, but an informative warning messages. It means that the Windows line endings are being changed to Unix line endings. Not doing so, may make programs not working well in different operating systems. This is explained in the [Pro Git book link](https://www.git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_core_autocrlf) linked just above this setting on our lessons.
New PR #894 created.

PR #886, opened 2022-07-21
Set appropriate format for code block in 'Exploring History' lesson 

sadielbartholomew:
In the ['Exploring History' lesson under the heading 'Simplifying the Common Case'](https://swcarpentry.github.io/git-novice/05-history/index.html#simplifying-the-common-case), there is a code block which is formatted as a 'bash' command when it is clearly a (partial) 'output' instead, and the preceding text even states this:

> If you read the output of git status carefully, you’ll see that it includes this hint:

This PR changes the specification of the title and syntax highlighting scheme to render that block in the manner appropriate to its context.

I have raised a PR directly instead of reporting this as an issue since it is a trivial and uncontroversial change which is simpler to fix directly. As far as I could tell, there isn't another PR open which addresses the formatting deficiency in question.

Thanks!

### DIFF
--- a/_episodes/05-history.md
+++ b/_episodes/05-history.md
@@ -327,7 +327,7 @@ here's how Git works in cartoon form:
 > ~~~
 > (use "git checkout -- <file>..." to discard changes in working directory)
 > ~~~
-> {: .language-bash}
+> {: .output}
 >
 > As it says,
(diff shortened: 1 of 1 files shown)


kekoziar (approved):
verified change.

kekoziar:
@sadielbartholomew Thanks for the fix!

PR #888, opened 2022-07-29
Very minor wording tweaks to 08-collab

edlowther:
Have switched "its" to "her" for consistency with other references to the Collaborator in the text, but would be equally happy with "their". 

Many thanks for developing this brilliant training material!


kekoziar (approved):
Thanks for catching this!

PR #894, opened 2022-09-19
IMPORTANT: Revert "Changed the Line Endings git config option so it works on Windows"

dpshelio:
Reverts swcarpentry/git-novice#880 - as it will make it hard to collaborate between different operating systems.

### DIFF
--- a/_episodes/02-setup.md
+++ b/_episodes/02-setup.md
@@ -70,7 +70,7 @@ For this lesson, we will be interacting with [GitHub](https://github.com/) and s
 > And on Windows:
 >
 > ~~~
-> $ git config --global core.autocrlf false
+> $ git config --global core.autocrlf true
 > ~~~
 > {: .language-bash}
 > 


kekoziar:
@dpshelio The line ending has two lists of commands based on operating systems.  The line which was changed was for Windows only, and was verified with the most current versions of Windows and Git-Bash.  

Please list the operating systems and the error messages received.  Thanks.

kekoziar:
I'm going to close this since 

> And on Windows:
>
> ~~~
> $ git config --global core.autocrlf false
> ~~~

this has been verified as the correct settings for Windows, and any changes to the `config core.autocrlf` command section specific to OSs other than the most current of Windows will require edits outside of what was changed in this PR.

dpshelio:
There are no error messages. It's a setting recommended to ease cross operating system contributions. If you take a look at the description on the [Pro-git book](https://www.git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_core_autocrlf) linked in the paragraph above in the lesson you will see why we recommend using  `true` instead of `false`. The warning messages that such setting produces (as referred in #880) is informative that the files are being changed to Unix format, not an error message.

kekoziar:
@dpshelio Then I think the section needs more explanation, perhaps
> A warning message will appear when a source file is first added, but setting `core.autocrlf` to `true` will reduce incompatibilities when a project is being worked on across OS platforms.

dpshelio:
Maybe a larger explanation is better (though there's one in the [teaching notes](https://swcarpentry.github.io/git-novice/guide/index.html#teaching-notes) already). At the moment the lesson is giving the incorrect suggestion that will hinder 
 collaboration.
I taught git today, and I can't point my learners to the official material for now.

dpshelio:
@kekoziar - any new thoughts about reverting to how it was  before and my last comment? I'm sending soon the material to the learners I had last week.

kekoziar:
The warning message is what confuses everyone.  I think a simple explanation - short and to the point - is the best.

dpshelio:
At the end (a year later) we got it right! #969 :wink:

PR #899, opened 2022-10-20
urgent: fix workflows

zkamvar:
This pull request updates the workflows for this lesson.

There are two items that are changed:

1. r-lib/actions/setup-r now uses `@v2` instead of `@master` as the default tag
2. the `set-output` GHA workflow command has been updated as it was deprecated.

see https://github.com/carpentries/styles/issues/641 for details

If you have any questions, contact @zkamvar

### DIFF
--- a/.github/workflows/template.yml
+++ b/.github/workflows/template.yml
@@ -127,7 +127,7 @@ jobs:
         id: check-rmd
         working-directory: lesson
         run: |
-          echo "::set-output name=count::$(shopt -s nullglob; files=($(find . -iname '*.Rmd')); echo ${#files[@]})"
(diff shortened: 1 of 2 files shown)


kekoziar (approved):
git-novice is a non-R lesson, so this doesn't impact the lesson build.  I checked the PR locally, and everything renders as normal.  Approved.

zkamvar:
> git-novice is a non-R lesson, so this doesn't impact the lesson build.

True, but the bug does affect PR statuses (e.g. https://github.com/swcarpentry/git-novice/pull/904) as detailed in the Impacts section of https://github.com/carpentries/styles/issues/641

PR #901, opened 2022-11-07
Remove trailing period from commit message

wjp:
This makes the commit message consistent with the others in this lesson. It also follows the advice from
https://chris.beams.io/posts/git-commit/ , which is referred to as a guideline for commit messages.

### DIFF
--- a/_episodes/06-ignore.md
+++ b/_episodes/06-ignore.md
@@ -94,7 +94,7 @@ Let's add and commit `.gitignore`:
 
 ~~~
 $ git add .gitignore
-$ git commit -m "Ignore data files and the results folder."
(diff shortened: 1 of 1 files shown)


kekoziar (approved):
I'm going to approve this, but with the following commentary: technically, this is the commit subject, which is often used alone - and abstractly - to describe the commit. If the commit subject or message has sentences, it should use proper punctuation. There is no need to restrict a commit to only N number of characters. And, while Chris Beams has been referred to on commit messages, there are other, more current, theories and suggestions on commit messages.  The point is to have them be meaningful and useful. if punctuation allows that, then it should be included.

PR #904, opened 2022-11-15
Clarify to not commit proposed change

lexnederbragt:
A learner missed this and had already committed the change before testing `git checkout`. This change intends to help avoid this mistake.

### DIFF
--- a/_episodes/05-history.md
+++ b/_episodes/05-history.md
@@ -495,8 +495,8 @@ moving backward and forward in time becomes much easier.
 >
 > `git checkout` can be used to restore a previous commit when unstaged changes have
 > been made, but will it also work for changes that have been staged but not committed?
-> Make a change to `mars.txt`, add that change, and use `git checkout` to see if
-> you can remove your change.
+> Make a change to `mars.txt`, add that change (but don't commit it yet),
+> and use `git checkout` to see if you can remove your change.
 {: .challenge}
 
 > ## Explore and Summarize Histories


kekoziar:
@lexnederbragt Having the challenge written in such a way that a learner never misses it, defeats the purpose of the challenge. Did they miss it because it's unclear, or because the learner didn't quite grasp the difference between a change being staged and committed?  I'd rather not write in such a way that it takes into account each mistake a learner might make.  

That said, this would probably be best clarified with a solution to the challenge.  Are you willing to write up a good solution instead, taking into consideration the common mistakes learners might make with this?

lexnederbragt:
I see your point and agree that my change would be making the challenge less, well, challenging. 
I like the idea of adding a solution and will submit one soon.

lexnederbragt:
PR updated with a proposed solution. Please review!

kekoziar (commented):
What version git are you running?  My version is prompting with `git restore` not `git checkout`

lexnederbragt:
I agree that recent versions of git use `git restore` to restore files instead of `git checkout`. However, the lesson is based on older versions of git, and uses `git checkout`. So, this PR needs to use the current lessons convention.

See also discussion in #691 on updating the git lesson to version of git that uses `git restore` (and `git switch`).

kekoziar (approved):
Looks good.  I made some minor changes to make the input/output in the solution standardized with the rest of the lesson, and added clarification in the challenge prompt. 

Thank you for submitting this solution!

PR #911, opened 2023-02-07
Add source code archive challenge to open science episode

joenio:
This new content was made by SWH ambassadors during the SWH symposium 2023, thanks to: Ludovic, Antoine, Guillaume, Maxence, Oceane, Andreas.

### DIFF
--- a/_episodes/10-open.md
+++ b/_episodes/10-open.md
@@ -120,6 +120,20 @@ by acting as a shareable electronic lab notebook for computational work:
 > approach for your current project and explain why.
 {: .challenge}
 
+> ## How to Archive my Source Code?
+>
+> Source code hosting sites come and go: established hosting sites such as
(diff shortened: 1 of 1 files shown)


kekoziar (changes requested):
Thank you for sharing information shared by Software Heritage (SWH) ambassadors during the Software Heritage (SWH) symposium 2023.  I agree that the open science lesson needs to be expanded upon. I like the idea of adding the call-out specifically on how to publish/archive code in a git repository.  However, the section should be expanded as a list to include current options, including Zenodo, which is supported by CERN, and shouldn't be going away anytime soon (and is the current preferable way to publish/archive git repositories).  

As a reference, there are many repositories which archive source code https://www.re3data.org/search?contentTypes[]=Source%20code

I will leave this PR open for now; however, know that if no changes are made which can be merged prior to the migration to the new infrastructure, it will be auto-closed as part of the process.

Issue #891, opened 2022-09-01
Recommended comments to clarify sections


SARAHHETTEMA:
I noticed a few areas where comments in the code could help clarify next steps and avoid running into error messages:

In the directories section (https://swcarpentry.github.io/git-novice/04-changes/index.html#directories):  there is a code block "git add <directory-with-files>". I think it would be helpful to add a comment such as "#use this general form to add all files in the directory at once" or remove the code formatting. I find it can be confusing to have a code chunk that is not supposed to be executed in the lesson without some additional commentary that the line of code is just example formatting. 

In the Ignoring things episode (https://swcarpentry.github.io/git-novice/06-ignore/index.html#overview): for the "nano .gitignore" line. I would add a comment to the output section that you need to edit the nano file with the "*.dat results/" text then run git status. I think either a comment next to the nano line or a line before the code chunk would clarify and prevent folks from getting stuck on that section.

emcaulay:
HI @SARAHHETTEMA, 

These are really helpful suggestions! Thank you for making them. I think both suggestions are worth incorporating into the lesson. I might work with the lesson maintainers to see if I can draft some revisions.

PR #912, opened 2023-02-08
Update 06-ignore.md according to issue #891

SvenMarcus:
Issue #891 proposed adding some additional comments on the creation of the `.gitignore` file to clarify that learners need to add the ignore patterns to the files themselves.

### DIFF
--- a/_episodes/06-ignore.md
+++ b/_episodes/06-ignore.md
@@ -48,10 +48,17 @@ What's worse,
 having them all listed could distract us from changes that actually matter,
 so let's tell Git to ignore them.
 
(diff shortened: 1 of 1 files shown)


kekoziar (changes requested):
Part of this is an approved recommendation; however, based on feedback from #710, it would be clearer if the change in line 51 is something similar to 
> We do this by creating a file in the root directory of our project called .gitignore. Then we'll add lines specifying which files we'd like to ignore in our project.

kekoziar (changes requested):
Changes starting on line 55 go against the standard of bash input/output used in the lesson.  Please remove.

PR #916, opened 2023-04-27
update zenodo json for lesson release

tobyhodges:
updating the `.zenodo.json` with metadata about contributors since last release, in preparation for the infrastructure transition.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -80,7 +80,8 @@
       "orcid": "0000-0003-3627-5340"
     },
     {
-      "name": "Nima Hejazi"
+      "name": "Nima S Hejazi",
+      "orcid": "0000-0002-7127-2789"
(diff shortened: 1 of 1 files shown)


kekoziar (commented):
Nima's listing under creators is missing the middle initial and orcid. Is the json file automatically created?  I'm wondering if it's okay to merge, or if that should be corrected.  I'm not sure of the policy on the contributors file.

tobyhodges:
Great catch, @kekoziar, thanks. The files are automatically created, though for reasons I admit I do not fully understand, the Editors list is generated slightly differently from the Creators list, which can result in people's details appearing differently in the two lists. I have updated to make Nima's details consistent.

kekoziar (approved):
looks good.

PR #927, opened 2023-06-05
change the use of git checkout to git restore

kleuveld:
I noticed that my version of git ( (2.36.1.windows.1) doesn't prompt the user to use git checkout to restore files. It uses git restore. This is much safer, and can't lead to nasty stuff like detached heads, so probably better for learners. I went ahead and changed all the references to git checkout to git restore.

What did I update:

- I changed git output to show what my version of git shows
-  I replaced all instances of git checkout with the relevant git restore command, including in exercises. 
-  I changed two pictures to use git restore. However, I don't have a SVG editor, so it looks fairly ugly. So that probably needs some work still.


kekoziar (dismissed):
Thanks for this needed update.  You missed some instances of `checkout` in this particular episode.  Can you update your PR?

kleuveld:
Hi, 

Sorry for the slow reply, I was on a holiday, so it got to the bottom of the pile. I have added a new commit that should change all the remaining instances of checkout to restore.

Cheers,
koen

peterjc:
See #691

martinosorb:
@kekoziar are you happy with the changes now? I'm happy to resolve the conflict myself, it's minor.

Issue #824, opened 2021-07-25
SSH (and PAT) In-Depth Supplemental Episode


kekoziar:
It has been decided (see https://github.com/swcarpentry/git-novice/issues/778) that minimum SSH is in the regular lesson, and a slightly advanced (including commands that are usually included in most tutorials for cloud repositories) will be added as a supplemental episode.

Since many people in the community are interested in contributing to this, and it might take some time to implement, we will first pursue this effort in a separate branch and later merge this branch into the main `gh-pages` branch.

This special branch has been created and is called `ssh-adv`. We, therefore, ask anyone who is interested in contributing to this effort to submit Pull Requests against this `ssh-adv` branch.

### How to contribute
Start by fetching and checking out the `ssh-adv` branch into your current repository

```shell
git fetch https://github.com/swcarpentry/git-novice.git ssh-adv
git checkout ssh-adv
```

Now, you can start working on it as usual:

```shell
git checkout -b my-changes
# ... make changes...
git add
git commit
git push <your-fork>
```

... and submit a PR! When you submit a pull request on changes to this supplemental SSH episode (15-supplemental-ssh.md), make sure to select `ssh-adv` branch of this repository as the base branch.  Only changes to that episode will be committed to the `ssh-adv` branch!

That's it!

_We hope to see **your** Pull Request soon_!

ostueker:
Hello Carpenter's,

I wrote up some additional content for this supplemental SSH lesson.  Let me know what you think of this.

I'm also thinking of covering the following:
* Using the macOS Keychain to store the SSH key.  The benefit for macOS users is that they don't need to start the `ssh-agent` every time but only need to add the key once with `ssh-add -K ~/.ssh/id_ed25519`.  Then the key should be automatically unlocked whenever they login to their account or on first use during a session. (I'm not a mac-user myself but several of my colleagues are so I will make sure that the information is correct)
* On Windows, users can install the PuTTY SSH client (or WinSCP or MobaXterm that bundles this) and then convert the SSH-key to the PuTTY format and load it into the "Pageant" SSH agent. Finally they just need to set an environment variable to use the plink.exe instead of ssh. Then it works very well with Git-for-Windows, but not with WSL or WSL2.
* Linux users can just rely on the SSH-agent or Keychain for their desktop.

I hope to have enough time to write that up early next week.

kekoziar:
[...]

PR #928, opened 2023-06-09
Removed reference to advanced SSH material which is still in developm…

longr:
Remove reference to advanced SSH material.

In reference to issue #824 this is a suggestion to remove the reference to supplementary material for advanced SSH.  The material is not yet written or available,  but this is not clear from the text, as such learners can be searching for material that is not there.

I have left the main callout as I think the explanation of SSH is useful.


kekoziar (approved):
@longr  Thanks for updating the lesson!

PR #932, opened 2023-06-21
Update 03-create.md

mwardeck:
Proposal to add a step to lesson 3 on line 85, asking the students to cd into and look at the contents of their .git directory.

Occasionally, git users will need to make configuration changes, so it may be beneficial for them to see where they can do this.

I am contributing this suggestion as part of my instructor training, thanks.

### DIFF
--- a/episodes/03-create.md
+++ b/episodes/03-create.md
@@ -82,6 +82,24 @@ including the tracked files and sub-directories located within the project's dir
 If we ever delete the `.git` subdirectory,
 we will lose the project's history.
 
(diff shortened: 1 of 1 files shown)


kekoziar:
I think this level of work is outside the scope of the lesson. Plus, it's far safer to use the `git config` commands to make configuration changes to an individual repository.  Just this week I had to assist a learner in opening their config file to remove an accidental character which caused a cascade of errors.

martinosorb:
I agree with Kat, it is best not to change configurations directly. I'm going to close this for now, let us know if you have further comments @mwardeck! And thanks for the contribution.

Issue #895, opened 2022-09-20
SSH vs HTTPS


Aariq:
I know this topic has been discussed at depth (#778), but I feel like it might be worth revisiting.  Almost every tutorial I encounter *except* the Carpentries recommends using HTTPS with a PAT for GitHub ([happygitwithr](https://happygitwithr.com/https-pat.html#https-vs-ssh), the [`usethis` R package](https://usethis.r-lib.org/reference/use_github.html), and the fact that it's the [GitHub default](https://stackoverflow.com/questions/11041729/why-does-github-recommend-https-over-ssh#:~:text=A%20GitHub%20repository%20is%20therefore%20more%20universally%20accessible%20using%20HTTPS%20than%20SSH.&text=SSH%20Keys%20do%20not%20provide,if%20your%20key%20is%20stolen.), to name a few). The reasons listed by these sources in favor of using and teaching HTTPS are:
- increased security (your PAT expires and can be easily deleted if someone were to get ahold of it)
- ease of setup (not all Windows machines have SSH, not all machines have [port 22 open](https://github.com/swcarpentry/git-novice/issues/864), firewalls, etc.)
- The PAT generated is also used for the GitHub API which also enables the use of many other tools that use the API under the hood in one setup step

The arguments I see in discussions in this repo in favor of SSH seem mostly philosophical:
- SSH is a widespread tool worth learning (although it's not like the concept of an API token is unique to GitHub)
- SSH is not owned by a private company (but GitHub certainly is, which is already being taught)
- SSH is platform agnostic (so is HTTPS??)

I just think this might be worth revisiting since many other git/GitHub tutorials strongly recommend HTTPS and it seems like the majority of sources think it's less likely to cause (technical) problems for students.

Aariq:
There were some very valid concerns about security of instructors showing their PAT or students having their PAT exposed (https://github.com/swcarpentry/git-novice/issues/778#issuecomment-883437322).  But most tutorials strongly recommend using a credential manager so that you are not keeping your PAT in any plain text file and instead use a git-supported [credential manager](https://git-scm.com/docs/git-credential).  Instructors should also delete the PAT they create for the course immediately after in the unlikely event that someone takes a screenshot and attempts to use it for malicious purposes.

kekoziar:
[...]

PR #934, opened 2023-07-17
Add HTTPS + PATs in advanced SSH section

Aariq:
I fleshed out the section on HTTPS and PATs a bit.  Something I didn't do that would probably be nice is add screenshots from GitHub.  I could add some, but I thought it would be better for continuity if they matched the user and repo name from earlier in the lesson.  I changed the bit about PATs being unique to GitHub in the beginning because they aren't.  Bitbucket also has an option to use HTTPS + PAT and you'll find plenty of other non-git services that use PATs for authentication. Hopefully I've explained that succinctly in the language I added.

### DIFF
--- a/_episodes/15-supplemental-ssh.md
+++ b/_episodes/15-supplemental-ssh.md
@@ -17,8 +17,7 @@ GitHub used to allow command line (CLI) authentication using only username and p
 uses HTTPS protocol.  However, this actually creates security risks, so, they removed 
 the CLI username/password option.  Instead, they kept two options which provide more
 robust security: SSH, which is a widely used protocol; and Personal Access Tokens (PAT), which uses 
-the HTTPS protocol, and is 
-specific to GitHub.  The minimum level of SSH to connect to GitHub is described in episode 7. 
+the HTTPS protocol.  The minimum level of SSH to connect to GitHub is described in episode 7. 
 
 This supplemental episode discusses SSH and key pairs in more depth and detail, 
 and provides details of advanced setup of SSH for GitHub.  It is included in this Git lesson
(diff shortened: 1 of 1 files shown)


Aariq:
btw, this is in response to this comment: https://github.com/swcarpentry/git-novice/issues/895#issuecomment-1563162295

kekoziar (approved):
Looks good.  Thanks for updating and expanding this supplemental episode.

kekoziar:
@Aariq Thanks for expanding this episode.  I have a couple of questions re the use of PATs.  When GitHub first required the increased security, I could only find the term "Personal Access Token" in relation to Microsoft products (GitHub, Azure, Microsoft, but mostly GitHub). Now, the terminology seems to have migrated to other git/version control/cloud collaboration services (Atlassian, GitLab, and some smaller providers).  Have you seen PATs in widespread use?  Are their implementation in other services the same as in GitHub?   

Thanks!

Aariq:
I have not seen the term "Personal Access Token" / PAT in widespread use, but the concept and terminology is closely related to other kinds of access tokens that _are_ widely used for authentication via HTTPS—e.g. API tokens, bearer tokens, etc. To be honest, it's not clear to me how exactly PATs differ from any other kind of access token. From a user standpoint, they seem to function about the same and the skills gained by learning to use GitHub PATs are very transferrable, IMO. Even if "Personal Access Tokens" are not commonly used, the use of access tokens to authenticate via HTTPS is _quite_ common (e.g. for some data APIs) and is a useful skill for researchers.

Aariq:
GitLab appears to have many more [types of tokens](https://docs.gitlab.com/ee/security/token_overview.html) including Personal Access Tokens, Group Access Tokens, Project Access Tokens, etc. But it doesn't actually look like they are used for [authenticating with git](https://docs.gitlab.com/ee/user/ssh.html).  [Bitbucket appears to use PATs](https://confluence.atlassian.com/bitbucketserver072/personal-access-tokens-1005335924.html) the same way as GitHub—you can use them when prompted for a password by git

Aariq:
Yeah, that makes sense to me. How's this look now?

kekoziar:
> Yeah, that makes sense to me. How's this look now?

Looks good.  Thanks!

PR #935, opened 2023-07-19
Remove duplicate caption, add alt text

### LINKED ISSUE #929: Caption appears twice
The caption/credit for "Piled Higher and Deeper" appears twice with the corresponding comic image on the [Git basics page](https://swcarpentry.github.io/git-novice/01-basics.html)

bencomp:
This fixes #929 and adds a description of the comic as alt text.

The title of the comic is *"FINAL".doc*; "Piled Higher and Deeper" is the name of the publisher.

### DIFF
--- a/episodes/01-basics.md
+++ b/episodes/01-basics.md
@@ -22,9 +22,7 @@ to keep track of what one person did and when.
 Even if you aren't collaborating with other people,
 automated version control is much better than this situation:
 
-!["Piled Higher and Deeper" by Jorge Cham, http://www.phdcomics.com](fig/phd101212s.png)
-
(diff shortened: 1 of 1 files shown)


kekoziar (changes requested):
Thanks for adding the alt-text to make lesson more accessible.

The title of that particular strip is "[notFinal.doc](https://phdcomics.com/comics/archive_list.php)"

bencomp:
Ah yes, I see: in the archives and in title of the page (that I see in my browser tab), it is indeed called "notFinal.doc". I had looked at the main header at the top of the strip.

kekoziar (approved):
Looks good.  Thanks for updating it!

PR #940, opened 2023-08-09
add link to conflict in glossary

maneesha:


kekoziar (changes requested):
Thanks for adding the link, but it goes to a page that doesn't exist.

maneesha:
Ugh, this is what I get for trying to put in changes during the workshop. Clearly I need to wait until I can focus.

martinosorb:
Hi @maneesha , I'm a bit confused on what is going on here -- the link still seems not to work for me. I believe the previous version with `html` was correct, perhaps the mistake was in `..learners` instead of `../learners`? I think you know better than me how the Workbench links work. Could you check again please?

kekoziar (approved):
Thanks for updating the syntax. Link works on local render.

PR #942, opened 2023-08-09
clarify language losing your work

maneesha:
Where it says nothing is lost unless you work hard at it, that could be read as the thing you worked hard on.  This clarifies that you have to work really hard to destructively lose all your changes.


martinosorb:
This makes sense. Thank you!

PR #943, opened 2023-08-09
link directly to section on default branch name

maneesha:
Makes link to go the section on default branch name rather than just to the setup episode.


kekoziar (approved):
Thank you for updating the link to the section.

PR #944, opened 2023-08-09
link to episode on hosting

maneesha:
In the remotes section we say we'll review the pros/cons of hosting services in a later episode.  This PR adds the link to that later episode.


kekoziar (changes requested):
Thanks for making the lesson easier to navigate by adding links.  But, the url syntax is supposed to be to a `.md` file

maneesha:
Ugh, this is what I get for trying to put in changes during the workshop. Clearly I need to wait until I can focus.

kekoziar (approved):
thank you for improving the episode

PR #947, opened 2023-08-10
note to go to settings in your repo

maneesha:
Add language to make it clear that you should go to settings in your repo, as otherwise it could be interpreted as looking for settings in the GH home page.


kekoziar (approved):

kekoziar (approved):
Thank you for your contribution. I took the opportunity to update the alt-text of the image, since those using screen-readers or relying on image alt-text in addition to the lesson text would likely not find the location to update collaborators in the settings.

PR #951, opened 2023-08-10
"Code" looks like a button, not a tab

maneesha:


kekoziar (approved):
Thank you for updating the lesson to reflect the most recent github UI.

PR #971, opened 2023-09-15
Remove note about ssh keys

lexnederbragt:
Unless I am mistaken, using ssh keys is now a requirement, which makes  this note outdated.


kekoziar (approved):
Thank you. This artifact was missed during the original adoption of ssh keys into the lesson.

Issue #903, opened 2022-11-11
Replace .dat file-ending with similar sensible ending


fiveop:
The material uses files that end in `.dat`, spoken “dot dat”. With unclear pronunciation or with bad acoustics this can be misheard as “dot dot” at some point and properly understood at another, which can then lead to confusion: Files are ending in `.dot` but .gitignore contains `*.dat`. (Based on recent experience.)

We could replace the file ending `.dat` with something distinct and recognizable, for example `.csv`, throughout the material.

kekoziar:
Changing it to `.csv`  sounds like a good idea.  Especially since `.dat` is often a data file that program developers will encounter, but not researchers (who are more likely to attend these lessons).

Feel free to submit a PR.

PR #972, opened 2023-09-18
Replace .dat with .csv

fiveop:
In the target audience .csv is more widespread file extension than .dat. In
addition, the spoken “dot dat” has led to confusion in that past (when being
understood as “dot dot”).

Solves #903


kekoziar (approved):

kekoziar (approved):
Thank you for updating the lesson!

PR #977, opened 2023-11-14
Include reference to git installation from chapter 2

unode:
For any reader/learner that may have missed the Summary and Setup section of the lesson, gently guide them to the corresponding page and the installation instructions section before introducing any command. 

This PR adds a callout and link to the git installation section just before the first `git command` in chapter 2.

[Click here for a preview of the change](https://unode.github.io/git-novice/02-setup.html#install-git-first)

### DIFF
--- a/episodes/02-setup.md
+++ b/episodes/02-setup.md
@@ -17,8 +17,14 @@ exercises: 0
 
 ::::::::::::::::::::::::::::::::::::::::::::::::::
 
+:::::::::::::::::::::::::::::::::::::::::  callout
+
+## Install Git first
+
 If you got this far but haven't installed Git yet,
-[follow these installation steps first](../learners/setup.md#installing-git).
+[follow these installation steps](../learners/setup.md#installing-git).
+
+::::::::::::::::::::::::::::::::::::::::::::::::::
 
 When we use Git on a new computer for the first time,
 we need to configure a few things. Below are a few examples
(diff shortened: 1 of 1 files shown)


martinosorb:
Sounds very reasonable and I don't see why we shouldn't include it. Thanks for the PR!

martinosorb:
For reference -- this merge has been reverted due to lack of agreement among maintainers.

unode:
Can a case still be made to have a reference to the setup in some form in this chapter?

The main motivation is to guide self-learners.

As the opening text in the lesson reads like a fiction work, we have noticed that learners often skip it, arriving to the first command without having git correctly installed.

An alternative to the callout proposed, could be to provide a some info to troubleshoot the eventual "git: command not found" if installation wasn't done or wasn't done properly (e.g. git not added to PATH).
This was the error most learners that skipped the setup ran into. Most of them were still able to launch a command-line (cmd.exe on Windows and Terminal in Mac/Linux).

kekoziar:
@unode While many Carpentries lessons can be used for self-learning, at their core, they are still designed to be led by a knowledgeable instructor.  We have to be careful with scope creep, and self-learners vs a workshop lesson is one of the areas which we need to be careful about. If we imagined all of the scenarios where self-learners may need assistance, it would be a very different (and large) lesson. 

That said, there is an approved solution that checks if learners (self or workshop) have installed git, by introducing `git status` at the beginning.  It's referenced in this [issue](https://github.com/swcarpentry/git-novice/issues/685#issuecomment-1528095979).

Feel free to include a PR with the proposed solution, or if you want we can workshop the wording here or in the original issue.

PR #980, opened 2023-11-21
Stress the importance of `git status` (#685)

### LINKED ISSUE #685: Episode 3 suggests to use git status before initializing a repo, but does not show that

jdvorak001:
Closes #685

Based on the discussion in the issue, I'm limiting the changes to just giving a bit of emphasis on the moment the `git status` command is first used.

### DIFF
--- a/episodes/03-create.md
+++ b/episodes/03-create.md
@@ -95,8 +95,7 @@ $ git checkout -b main
 Switched to a new branch 'main'
 ```
 
-And now we will use what I think is the most important Git command: `git status` reports on the status of our project.
-It tells us the status of our project, and better, a list of changes in the project and options on what to do with those changes.
+We can now start using one of the most important git commands, which is particularly helpful to beginners. `git status` tells us the status of our project, and better, a list of changes in the project and options on what to do with those changes. We can use it as often as we want, whenever we want to understand what is going on.
 
 ```bash
 $ git status
--- a/episodes/03-create.md
+++ b/episodes/03-create.md
@@ -95,8 +95,8 @@ $ git checkout -b main
 Switched to a new branch 'main'
 ```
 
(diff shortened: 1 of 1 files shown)


kekoziar (commented):
I'm not sure it's the same as `ls`. Maybe, instead of _It's the Git equivalent of listing a directory in a shell._ it can be _"It tells us the status of our project, and better, a list of changes in the project and options on how to resolve those changes"_

jdvorak001:
Ok, I see it's a bit imprecise to say _the Git equivalent of ..._

kekoziar (approved):
I like it.  Thanks for the quick change.  I tagged another maintainer for review.

martinosorb:
I like the idea of adding more emphasis on this command at that point. However, I think the phrasing is a bit off:

- _And now we will use what I think is the most important..._: I think nowhere in Carpentries materials the lessons use "I" -- who is "I" here?
- `git status` _reports on the status of our project. It tells us the status of our project..._ there is a clear repetition here

I would suggest something like _We can now start using one of the most important git commands, which is particularly helpful to beginners. `git status` tells us the status of our project, and better, a list of changes in the project and options on what to do with those changes. We can use it as often as we want, whenever we want to understand what is going on._

martinosorb (changes requested):
See above.

jdvorak001:
@martinosorb, your formulation is very good. I committed it with you as co-author.

martinosorb (approved):

martinosorb:
Thank you so much!

PR #988, opened 2024-04-17
Add descriptive alt text to all images

multimeric:
My sense is that some of the current image alt texts are not descriptive enough. For example "The Git Commit Workflow" is more of a title than a description. I have generally attempted to follow the following principles: https://carpentries.org/blog/2022/11/accathon/.

### DIFF
--- a/episodes/01-basics.md
+++ b/episodes/01-basics.md
@@ -38,20 +38,20 @@ think of it as a recording of your progress: you can rewind to start at the base
 document and play back each change you made, eventually arriving at your
 more recent version.
 
-![](fig/play-changes.svg){alt='Changes Are Saved Sequentially'}
+![](fig/play-changes.svg){alt='A diagram demonstrating how a single document grows as the result of sequential changes'}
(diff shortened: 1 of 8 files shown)


pgmccann:
I like these, and think they are generally a big improvement, but I have a concern about the accessibility of text which references "the previous/earlier diagram". I believe it is generally not good practice to refer to an element of a webpage by its position alone.

martinosorb:
Thank you for the PR @multimeric. If you can think of a meaninful way to change your alt texts according to @pgmccann's comment, that would be great. Otherwise I will merge as-is, which is already a clear improvement.

multimeric:
Okay, I'll try to think about how I might explain these diagrams without referring to earlier diagrams, but I think it might end up being quite verbose.

martinosorb:
Thank you @multimeric.
@kekoziar I will merge unless you have objections.

multimeric:
Sure, this is all good from my end.

PR #992, opened 2024-07-02
Further warnings in rm exercise

### LINKED ISSUE #965: Add note about how rm from command line is irrecoverable

In the [section about removing the `.git` file](https://swcarpentry.github.io/git-novice/03-create.html#correcting-git-init-mistakes) if we accidentally did a `git init`, there is a note to use with caution.  In emphasizing that caution, consider adding a line about how the `rm` command is irrecoverable -- items do not end up in your trash/recycle bin but are gone forever.  

https://swcarpentry.github.io/git-novice/03-create.html#correcting-git-init-mistakes

martinosorb:
Fixes #965, additionally removing a 'just' and clarifying we don't want to remove the whole subdirectory.

### DIFF
--- a/episodes/03-create.md
+++ b/episodes/03-create.md
@@ -166,7 +166,7 @@ fatal: Not a git repository (or any of the parent directories): .git
 ## Correcting `git init` Mistakes
 
 Wolfman explains to Dracula how a nested repository is redundant and may cause confusion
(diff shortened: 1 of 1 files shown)


erinmgraham (approved):
Challenge now notes rm is irreversible.

PR #994, opened 2024-07-02
Remove 'AFK'

### LINKED ISSUE #957: "AFK" may not be a recognizable term to everyone

In the [typical work session](https://swcarpentry.github.io/git-novice/09-conflict.html#a-typical-work-session) exercise, the last step is to "celebrate" with "AFK."
I am assuming this means "away from keyboard" but everyone may not know this, especially because  everything else in that exercise is an actual shell/git command and it may cause confusion to think that "AFK" may also be a command they should know.

https://swcarpentry.github.io/git-novice/09-conflict.html#a-typical-work-session

martinosorb:
Fixes #957.


erinmgraham (approved):
AFK removed AFAIK 😂

PR #995, opened 2024-07-02
Explain no feedback in typing passphrases

### LINKED ISSUE #956: Note that there is no response when typing password/pass phrase/etc in your terminal

In the [part of the git lesson about setting up an ssh key](https://swcarpentry.github.io/git-novice/07-github.html#create-an-ssh-key-pair) we should note that when typing in passwords in terminal, there is no feedback (ie your cursor doesn't move, you don't see your display changing, etc.).  We had several learners in a workshop today who thought their computer had frozen and was being non-responsive.

https://swcarpentry.github.io/git-novice/07-github.html#create-an-ssh-key-pair

martinosorb:
Fixes #956.

### DIFF
--- a/episodes/07-github.md
+++ b/episodes/07-github.md
@@ -199,7 +199,10 @@ Created directory '/c/Users/Vlad Dracula/.ssh'.
 Enter passphrase (empty for no passphrase):
 ```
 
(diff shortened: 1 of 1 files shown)


erinmgraham (approved):
Note added for no visual feedback.

PR #996, opened 2024-07-02
Update Citation

### LINKED ISSUE #945: Link and text to SWC Citation file do not match

In the [Citation section](https://swcarpentry.github.io/git-novice/12-citation.html) we see:

> You may want to include a file called CITATION or CITATION.txt that describes how to reference your project; the [one for Software Carpentry](https://github.com/swcarpentry/website/blob/gh-pages/CITATION) states:

followed by text that looks like it was copied and pasted from the referencing URL but it does not match what is currently there (or in the history).   

https://swcarpentry.github.io/git-novice/12-citation.html

martinosorb:
Fixes #945.
I copied the current version. Since it has not been updated in 8 years, it feels safe enough.

### DIFF
--- a/episodes/12-citation.md
+++ b/episodes/12-citation.md
@@ -23,30 +23,18 @@ Carpentry](https://github.com/swcarpentry/website/blob/gh-pages/CITATION)
 states:
 
 ```source
-To reference Software Carpentry in publications, please cite both of the following:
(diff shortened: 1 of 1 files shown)


erinmgraham (approved):
Content changed to match the more recent Carpentry text

PR #997, opened 2024-07-02
Fix "second big question"

### LINKED ISSUE #946: A second point is introduced without a first point
### What is the problem?

The [Hosting episode](https://swcarpentry.github.io/git-novice/13-hosting.html) begins with "The second big question for groups..." but does not follow what the first big question might be.  

https://swcarpentry.github.io/git-novice/13-hosting.html

martinosorb:
Fixes #946.


erinmgraham (approved):
Fix refers to first question directly

PR #1000, opened 2024-07-03
Fix syntax with alt text

### LINKED ISSUE #999: Some images have alt text displayed in lesson
### What is the problem?

{alt=‘Clicking the “Copy to Clipboard” button on GitHub to obtain the repository’s URL’}
is shown as part of the text. There are several places where this is the case. I believe it was introduced in a recent PR. Will look into this asap.

e.g. https://swcarpentry.github.io/git-novice/aio.html#connect-local-to-remote-repository

martinosorb:
Fixes #999.

### DIFF
--- a/episodes/07-github.md
+++ b/episodes/07-github.md
@@ -68,7 +68,7 @@ which looked like this:
 
 Now that we have two repositories, we need a diagram like this:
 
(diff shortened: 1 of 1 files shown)


martinosorb:
Merging because this is a bug in the display of images and I think it's better to address it quickly.

PR #1001, opened 2024-07-03
Mention both git log and git show (#875)

### LINKED ISSUE #875: Episode 5: Typo in instructions for referring to commits
In Episode 5, this is the wording for how to refer to commits by ID:

> We can also refer to commits using those long strings of digits and letters that `git log` displays.

But `git log` has not been called thus far during the episode. I recommend amending this to `git show`, as that command is the most recently run and does contain the alphanumeric string mentioned.

If you are OK with this change, I can put in a PR for this on Thursday.

dicksonlaw583:
In episode 5, mention that both `git log` and `git show` display IDs that can identify commits.

Resolves #875

### DIFF
--- a/episodes/05-history.md
+++ b/episodes/05-history.md
@@ -122,7 +122,7 @@ while `HEAD~123` goes back 123 commits from where we are now.
 
 We can also refer to commits using
 those long strings of digits and letters
-that `git log` displays.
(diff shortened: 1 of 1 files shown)


martinosorb:
Looks good to me.

PR #1010, opened 2024-08-06
Change settings side menu title

astroDimitrios:
I was looking at #990 and my account shows the same config options as the lesson but the `Account Settings` side menu subtitle is `Access` on mine. Hopefully this is the same for others and an easy change.


martinosorb:
I confirm it's called "Access". I will also close #990 as it seems outdated. Thank you!

Issue #960, opened 2023-08-10
Creating a new branch or renaming the default branch 


maneesha:
### How could the content be improved?

In the [Creating a Repository episode](https://swcarpentry.github.io/git-novice/03-create.html) we say we are going to rename the default branch.  However, command below (`git checkout -b main`) only creates a new branch called `main`; it does not change the default.  It's not clear here that you must follow the instructions in the linked git setup lesson to change the default branch name.  

### Which part of the content does your suggestion apply to?

https://swcarpentry.github.io/git-novice/03-create.html

kekoziar:
Thank you for the suggestion.  That part of the lesson is already in an individual repository.  _"Next, we will change the default branch to be called main."_ is in relation to the repository just created, so the default branch is now `main` in that particular repository.  I just tested it in a new repository, and once `main` is created, an error message is returned if trying to checkout master. 

```
$ git init
Initialized empty Git repository in C:/Users/vlad/repos/test/.git/

vlad@MS-NADF4Y2  ~/repos/test (master)
$ git checkout -b main
Switched to a new branch 'main'

vlad@MS-NADF4Y2 ~/repos/test (main)
$ git checkout master
error: pathspec 'master' did not match any file(s) known to git
```

do you have a suggestion to make it clearer?

maneesha:
The lesson says:

> Next, we will change the default branch to be called main. 

The instructions in this particular lesson do not do anything to change the name of the default branch.  Knowing that the default branch is likely already called `main`, instead of saying this is something we **will** do, we could say that we'll check to see what the branch name is, change it if needed, and then continue on.

kekoziar:
Since there are no commits on `master`, executing `git checkout -b main` effectively changes the name of the default branch to `main`. `master` can no longer be accessed.  So, the default branch of the repository is now `main` (or at least will be once a commit is added - technically at this point the repo doesn't have any branches, and executing `git branch -a` will return nothing, but this is far beyond a novice lesson).

I appreciate your suggestion to change the text to say we'll check the branch name; feel free to provide example text here if you want to workshop this, or submit a PR. 

Also, I recognize that you have elevated privileges in the Carpentries repositories, but please submit PRs from your own forked repo instead of creating new branches in this lesson's repository.  Unless there is a new workflow that I'm unaware of?  @ErinBecker ??

martinosorb:
[...]

PR #1011, opened 2024-08-06
Remove checkout branch main section in episode 3

### LINKED ISSUE #983: Do not mention branches while creating the first repository?

The setup episode includes configuring the git environment and setting the deafult branch to 'main'. The lesson for creating the first repository uses `git checkout -b main` again, to make sure that we are in the 'main' branch.

This might not be necessary since we already do the configuration step in the previous episode. This forces us to talk about branches prematurely in a context where we want to focus on creating new repositories and the `.git` directory. This might confuse some learners. If we mention branches, we also need more time to explain them, which would shift the focus of this episode. 

https://swcarpentry.github.io/git-novice/03-create.html

astroDimitrios:
Closes #983 and #960

Removes the section of episode 3 which asks you to checkout the main branch since this should be the default if you followed the steps in the previous episode.

### DIFF
--- a/episodes/03-create.md
+++ b/episodes/03-create.md
@@ -82,19 +82,6 @@ including the tracked files and sub-directories located within the project's dir
 If we ever delete the `.git` subdirectory,
 we will lose the project's history.
 
-Next, we will change the default branch to be called `main`.
-This might be the default branch depending on your settings and version
-of git.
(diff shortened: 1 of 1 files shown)


martinosorb:
I'm happy with this, thanks @astroDimitrios!

PR #1012, opened 2024-08-14
changes md symbol used for listing items

dpshelio:
This is something that I always forget and only remember when I'm teaching it and encountered as recently @krishnakumarg1984 reminded me.

Markdown accepts either `-` or `*` however, in this material, the `-`
causes confusion when looking at diffs as you'd get things like:

```diff
 - avocado
-- lemon
+- lime
```

making the `-` from the diff misunderstood with the `-` of the list.

### DIFF
--- a/episodes/04-changes.md
+++ b/episodes/04-changes.md
@@ -202,9 +202,9 @@ $ cat guacamole.md
 
 ```output
 # Ingredients
-- avocado
-- lemon
-- salt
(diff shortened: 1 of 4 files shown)


tobyhodges:
I am a big fan of this suggested change! 🎉

dpshelio:
It's included into #1013 to avoid conflicts.

PR #1003, opened 2024-07-03
Recipes draft

martinosorb:
I created this PR to check if  it was possible to merge the Recipes version without too much pain, and so we can discuss whether this is a good choice.
Before merging we have a number of TODOs:

- [x] Ask @dpshelio if there are any objections to using their material and how we can acknowledge them
- [x] Ask @swcarpentry/git-novice-es-maintainers how much would this affect the Spanish language fork and whether they are ok with updating their translation accordingly
- [x] As @swcarpentry/git-novice-maintainers, consider if this is what we want to do, as opposed to other choices
- [x] Under suggestion from @kekoziar, avoid cocktails and use something alcohol-free instead
- [x] Gather opinions from the CAC @swcarpentry/governance-committee

PR #1013, opened 2024-08-14
Modifies examples to include a single heading and changes md symbol used for listing items

dpshelio:
As per @tobyhodges comment on https://github.com/swcarpentry/git-novice/pull/1003#pullrequestreview-2196084226

But it also includes title to make the files consistent with their filenames. (diffs were also updated accordingly)

Update (brought from #1012)
This is something that I always forget and only remember when I'm teaching it and encountered as recently @krishnakumarg1984 reminded me.

Markdown accepts either `-` or `*` however, in this material, the `-`
causes confusion when looking at diffs as you'd get things like:

```diff
 - avocado
-- lemon
+- lime
```

making the `-` from the diff misunderstood with the `-` of the list.

### DIFF
--- a/episodes/04-changes.md
+++ b/episodes/04-changes.md
@@ -40,8 +40,9 @@ $ nano guacamole.md
 Type the text below into the `guacamole.md` file:
 
 ```output
-# Ingredients
-# Instructions
+# Guacamole
+## Ingredients
+## Instructions
 ```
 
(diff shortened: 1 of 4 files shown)


dpshelio:
Since the heading sections is something that's got already "semi-approved" I'm also pushing here the content from #1012 as it will produce conflicts otherwise.

PR #1014, opened 2024-08-25
Add missing periods in ep 05

tomeknd:
This is my first ever PR, so any feedback appreciated if applicable. Thank you for maintaining this awesome project!

### DIFF
--- a/episodes/05-history.md
+++ b/episodes/05-history.md
@@ -316,7 +316,7 @@ Answer 3 gets the version of `data_cruncher.py` from the commit *before* `HEAD`,
 what we wanted.
 
 Answer 1 results in an error. You need to specify a file to restore. If you want to restore all files
-you should use `git restore .`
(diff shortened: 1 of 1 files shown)


martinosorb:
Thank you for catching the missing periods and for taking the time to submit this PR. I have to point out that the first change would render as `git restore .`. which looks an awful lot like `git restore ..` which is a different command (referring to the parent directory `..` rather than the current directory `.`). So I would avoid the period there to avoid confusion...

tomeknd:
Makes sense, thank you for the feedback. If the other edit is to be applied, is there any further action needed on my part?

martinosorb:
I've updated it myself.
I will not merge for a while because we are waiting for a larger merge which I don't want to conflict with this. Thank you.

tomeknd:
Thank you.

PR #1017, opened 2024-09-18
Fix keypoint on GPL

fkohrt:
As the GPL does not mandate sharing, putting the whole derivative under the GPL license is only required if licensees decide to share it.


martinosorb:
After confirming this is correct through the GPL FAQs, I'm happy to merge. Thanks for the contribution.

PR #1018, opened 2024-10-02
Adds callout to introduce Markdown

pgmccann:
Borrows heavily from https://github.com/swcarpentry/python-novice-gapminder/blob/2aef01ba47cbf88e9c607f920b83b7dc59d7044e/episodes/01-run-quit.md. Requires update to varnish so that table within callout is styled properly.

I'm not fully across the new lesson platform, so I'm not entirely clear how to update varnish for local testing with my fork of this lesson.

### DIFF
--- a/episodes/04-changes.md
+++ b/episodes/04-changes.md
@@ -45,6 +45,56 @@ Type the text below into the `guacamole.md` file:
 ## Instructions
 ```
 
+:::::::::::::::::::::::::::::::::::::::::  callout
+
+## Markdown
+
+Jimmy and Alfredo are using [Markdown](https://en.wikipedia.org/wiki/Markdown)
+to write their recipes. It is a simple plain-text format for writing lists,
+links and other things that might go into a web page.
+
+Table: Showing some markdown syntax and its rendered output.
+
++---------------------------------------+------------------------------------------------+
+| Markdown code                         | Rendered output                                |
++=======================================+================================================+
++---------------------------------------+------------------------------------------------+
+| ```                                   | <p></p>                                        |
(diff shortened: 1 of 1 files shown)


martinosorb:
Hi @pgmccann, I have a few concerns: it's a large table that may be a bit bulky here. I am not sure if the "callout" actually works due to Varnish, as you say (I remember working on this on the python-novice-gapminder side). Lastly, there is no introduction to it: it should be put in context and explained why we are showing this table through an introduction to what markdown is.

pgmccann:
Hi @martinosorb, I don't disagree that it's a bit bulky. I've lifted the table from the Python lesson, and removed some of the content, but it is still unwieldy. I think the information content is appropriate to a callout - perhaps it should be formatted differently. I have included a short introduction:

> Jimmy and Alfredo are using [Markdown](https://en.wikipedia.org/wiki/Markdown) to write their recipes. It is a simple plain-text format for writing lists, links and other things that might go into a web page.

erinmgraham:
I had a go at rendering the table as a callout and it looks all right. There is also a division tag known as a spoiler 
ex_callout_spoiler_markdown_table
that is similar to the callout except it comes up as a hidden dialog box you need to click to open. Doing it this way hides the text but for some reason it's rendering with an orange character block for A1 level headings. I couldn't find any hidden characters in the markdown; it might have something to do with the spoiler css? I created a separate PR #1077 - please have a look and let us know you're thoughts.

martinosorb:
I like the idea of a Spoiler. The callout alone would be too distracting in the flow of the lesson. So the only problem is the orange thingy at the moment? Even so, we could go ahead.

PR #1019, opened 2024-10-03
Update overlooked references to "Vlad" 🙂

simonwiles:


martinosorb:
Thank you for noticing this!

PR #1030, opened 2024-10-23
fix: resolve #1025

### LINKED ISSUE #1025: Recovering Older Versions of a File mistake in excercise
### What is the problem?

Recovering Older Versions of a File exerciese in the lesson has an answer that does not correspond to the question. Correct solution says it is both 2 and 4 that are correct but explanation says that 2 is not what we want. My guess is this is just a number mismatch and the correct answer should either say 3 and 4 are correct or reorder the items in the question.

https://swcarpentry.github.io/git-novice/instructor/05-history.html

### LINKED ISSUE #1026: Mistake in "guacamole.md contains a single line, which we can see by running:"
### What is the problem?

The output clearly shows three lines.

https://swcarpentry.github.io/git-novice/instructor/04-changes.html

novica:
Had to redo this because I was not planning on doing more than one PR. But so it happens, I am doing more than one PR :).

Closes  #1026

### DIFF
--- a/episodes/04-changes.md
+++ b/episodes/04-changes.md
@@ -55,7 +55,7 @@ $ ls
 guacamole.md
 ```
 
-`guacamole.md` contains a single line, which we can see by running:
+`guacamole.md` contains three lines, which we can see by running:
 
 ```bash
 $ cat guacamole.md


martinosorb:
Note to self: it also says "1 insertion" in an output

PR #1031, opened 2024-10-23
fix: resolves #1027

### LINKED ISSUE #1027: Improve file creation explanation

In https://swcarpentry.github.io/git-novice/instructor/04-changes.html the text goes as follows:
```
Let’s create a file called guacamole.md that contains the basic structure to have a recipe.
```

Then it says:

```
But remember, the bash command to create or edit a new file 
```

Then it justi gives the command 

```
$ nano guacamole.md
```

None of this makes clear the actually running the above *creates* a file as the introductory sentence states. In fact, it does not. It merely opens the nano buffer. The file will be created upon saving.

I think this needs to be reworded. Maybe the file can be created using `touch`. That would be one of the usual ways to create an empty file in `bash`. If not, then explain that the file is not there until saved.

The lesson goes on to run `ls` but running `ls` before saving will not show a file, and there is nothing in the text about saving.

https://swcarpentry.github.io/git-novice/instructor/04-changes.html

novica:
Closes #1027 as discussed in https://github.com/swcarpentry/git-novice/issues/1027

### DIFF
--- a/episodes/04-changes.md
+++ b/episodes/04-changes.md
@@ -45,7 +45,7 @@ Type the text below into the `guacamole.md` file:
 ## Instructions
 ```
 
-Save the file and exit your editor. Next, let’s verify that the file was properly created by running the list command (ls):
+Save the file and exit your editor. Next, let’s verify that the file was properly created by running the list command (`ls`):
 
 ```bash
 $ ls
--- a/episodes/04-changes.md
+++ b/episodes/04-changes.md
@@ -31,7 +31,7 @@ Let's create a file called `guacamole.md` that contains the basic structure to
 have a recipe.
 We'll use `nano` to edit the file;
 you can use whatever editor you like.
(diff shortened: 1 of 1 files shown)


episodes/04-changes.md
```diff
@@ -45,7 +45,7 @@ Type the text below into the `guacamole.md` file:
 ## Instructions
 ```
 
-Let's first verify that the file was properly created by running the list command (`ls`):
+Save the file and exit your editor. Next, let’s verify that the file was properly created by running the list command (ls):
```
martinosorb, line 48:
```suggestion
Save the file and exit your editor. Next, let’s verify that the file was properly created by running the list command (`ls`):
```

The backticks on `ls` should not be removed.

martinosorb:
Other than the comment above, I'm happy to merge.

martinosorb:
Thanks!

Issue #687, opened 2019-08-20
Relocate directories and .gitkeep, Episode 4 to .gitignore in Ignoring Things, Episode 6


Tammoo:
I'd like to suggest that Directories and .gitkeep at the very end of Episode 4 - Track Changes be relocated to the beginning of Episode 6 - Ignoring Things and have Episode 6 renamed to Keeping and Ignoring Things.

Here's my reasoning:
The Episode 4 section on directories and .gitkeep seem to come out of nowhere at the end of the episode and there really isn't any sort of strong explanation as to how .gitkeep is use to track empty directories or why one might want an empty directory. There is an example of how to set up a directory and add files to it using touch, but this is also confusing due to the files not having extensions. It seems that the file can be opened in nano, but not saved with an extension such as .txt. So, we end up having a spaceships directory with two files that appear to have unknown extensions. In Episode 5 directories and .gitkeep are not mentioned at all which seems to cause a bit of a disruption between the end of Episode 4 - directories and .gitkeep and all of Episode 6 - .gitignore. The flow might be better if directories and .gitkeep are a part of the conversation with .gitignore. .gitignore is not the opposite of .gitkeep, but I think their roles in relationship to directories and files, makes more sense if they are in the same episode.

Thanks so much. 
Tami

kpegion:
I agree. .gitignore and .gitkeep have not yet been introduced in the Lesson and seem to come from nowhere at the end of Episode 4.  Up to this point Episode 4 flows very nicely and is easy to follow.  Since the lesson is designed with a git novice as the target learner,  I think these two files are introduced here in a way that could lead to confusion.

JoKeyser:
I agree as well, I tripped over this when reading the lesson from the start.

PR #1036, opened 2024-11-03
Remove reference to gitignore

erinmgraham:
I agree with issue #687 contributors that the reference to .gitignore in this callout comes out of nowhere so I'm removing it for now. 

However, this callout needs to be revisited to consider having it in this location in the lesson and the fact that it makes changes that might be missed. (Instructors pressed for time will selectively skip callouts and this may have downstream effects.)

Note that there are some unintended changes due to line endings or whitespace adjustments which I will aim to fix in the PR :-(


martinosorb:
Looks good to me, but I'm confused on why Git is marking every line of the file as a change. Is there a way to fix it? If it's about line endings, I'm worried this might mess up the markdown parser (although it's unlikely).

erinmgraham:
I was using RStudio to make the changes so I could build the pages and test before pushing. It was saving the changes with CRLF.  I changed the Line ending conversion setting under Tools | Global Options | Code | Saving to LR.
________________________________

martinosorb:
I noticed it's possible to get github to show the actual changes by clicking on the gear and selecting 'hide whitespace'.

Should we keep the part that says 'these files are not special and their sole purpose is to populate a directory so that Git adds it to the repository' perhaps?

PR #1038, opened 2024-11-03
Amend gendered language

### LINKED ISSUE #838: Amending Gendered Language in Episode 8
In Episode 8: [Collaborating](https://swcarpentry.github.io/git-novice/08-collab/index.html), there are many gendered pronouns that should be changed to better fit with the model of inclusivity that the Carpentries has built. 

> To accept access to the Owner’s repo, the Collaborator needs to go to https://github.com/notifications. Once there she can accept access to the Owner’s repo.

- The second sentence is a bit vague so could be rewritten as "Once there, the Collaborator can accept access to the Owner's repo." Though the thought is a bit redundant so it would read better if the second sentence was dropped all together.

> Next, the Collaborator needs to download a copy of the Owner’s repository to **her** machine. This is called “cloning a repo”.

> To clone the Owner’s repo into **her** Desktop folder, the Collaborator enters:

> The Collaborator can now make a change in **her** clone of the Owner’s repository, exactly the same way as we’ve been doing before:
- In all three instances _her_ should be changed to _their_ as to not introduce any potentially demotivating language particularly to gender non-conforming learners.

erinmgraham:
Closes #838

Replaced he/she/him/her/his/hers with they/them/their/theirs to be more inclusive.

### DIFF
--- a/episodes/02-setup.md
+++ b/episodes/02-setup.md
@@ -27,7 +27,7 @@ of configurations we will set as we get started with Git:
 
 On a command line, Git commands are written as `git verb options`,
 where `verb` is what we actually want to do and `options` is additional optional information which may be needed for the `verb`. So here is how
-Alfredo sets up his new laptop:
+Alfredo sets up their new laptop:
 
 ```bash
 $ git config --global user.name "Alfredo Linguini"
@@ -84,7 +84,7 @@ $ git config --global core.autocrlf true
 
 ::::::::::::::::::::::::::::::::::::::::::::::::::
 
-Alfredo also has to set his favorite text editor, following this table:
+Alfredo also has to set their favorite text editor, following this table:
 
 | Editor                                | Configuration command | 
 | :-----------                          | :------------------------------ |
@@ -119,7 +119,7 @@ If you want to save your changes and quit, press <kbd>Esc</kbd> then type `:wq` 
 
 Git (2.28+) allows configuration of the name of the branch created when you
 initialize any new repository.  Alfredo decides to use that feature to set it to `main` so
-it matches the cloud service he will eventually use.
+it matches the cloud service they will eventually use.
 
 ```bash
 $ git config --global init.defaultBranch main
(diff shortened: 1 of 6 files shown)


martinosorb:
Thanks Erin. To my ears, Alfredo sounds like a masculine name, and Jennifer like a feminine name. So using 'they/them' with these two is a bit unnatural and distracting to me. Personally, I would keep he and she for these. The changes make total sense when it comes to the Owner and Collaborator in Episode 8.
We could also change Jimmy to another female (or gender-unspecified) character in Episode 3 for more balance, or even change Alfredo's name, but I wouldn't use they with a clearly masculine name...

martinosorb:
Jules and Ren could be ok, although we lose the funnyness of Alfredo Linguini as a name. Any thoughts from @dpshelio?

ErinBecker:
Coming late to the party to add my own thoughts . . . 

I agree with the original issue raiser (@omahoneym) that we should use gender neutral pronouns when referring to a person of unknown gender (e.g. the hypothetical collaborator). 

However, when we are referring to a named person in our lesson (a "character"), that character has a gender, and that gender is whatever the character's creators decide. If I'm reading correctly that this lesson has three characters, I personally like the idea of one of those characters using they/them pronouns. 

In terms of whether the character who uses they/them pronouns should have a gender ambiguous name, I don't think that's necessary. I know many they/them pronoun users with gender coded names. 

My proposal would be the following: 
- Jenny (she/her)
- Alfredo (they/them)
- Jimmy (he/him)

dpshelio:
I like @ErinBecker suggestion. Use the three characters as she suggests, having Alfredo as They/Them. However, thinking of translations, I really have no clue how that would translate to other languages, but I suppose each one would be free to adapt it to their best approach to this. I wonder if this kind of decisions (and their reasoning) should be reflected in the lesson's files rather than buried within the issues so translators team can benefit from it (as well as instructors or non-native speakers). 

So, should we create a page with lesson development considerations? That could include links to issues that discuss the changes from the planets to the recipes, the name of the characters and their pronouns...

PR #1051, opened 2024-11-15
Add a missing $ in bash listing

JoKeyser:


martinosorb:
Thanks for noticing this, looks good.

JoKeyser:
@martinosorb , thank you for merging.
I notice that the resulting merge commit does not mention me as author, nor my commit message. I think the Git history would be better if both were mentioned. Is there a convenient way to do this (in the future)?

PR #1053, opened 2024-11-19
add a missing $ in bash listing

JoKeyser:
Hi, while reading, I found another missing dollar sign in a bash listing.


martinosorb:
Thanks.

PR #1057, opened 2024-12-09
A few corrections to episode 4

fiveop:
Commits:
- No. 1: I think the change makes the sentence easier to read.
- No. 2: Remove accidental word repetition
- No. 3: Remove contraction from prose (I noticed that his was done consistently throughout the document; feel free to remove this commit from the PR, if that was intentional.)

### DIFF
--- a/episodes/04-changes.md
+++ b/episodes/04-changes.md
@@ -30,7 +30,7 @@ $ cd ~/Desktop/recipes
 Let's create a file called `guacamole.md` that contains the basic structure of a recipe.
 We'll use `nano` to edit the file;
 you can use whatever editor you like.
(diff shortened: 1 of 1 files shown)


martinosorb:
Hi @fiveop, thanks for this. The first two changes are great and should be merged.
While I agree that there is no consistency in using or not using contractions in this whole lesson, we shouldn't change this in a single episode and leave the others the same. I would not want you to go through every file in the whole lesson, so I would suggest we leave it as it is from this point of view.

fiveop:
I removed the third commit.

martinosorb:
Thank you!

PR #1061, opened 2025-01-10
clarify code block that is not run and break out nano steps

### LINKED ISSUE #891: Recommended comments to clarify sections

erinmgraham:
Closes #891

Code blocks with 'bash' that are not can be confusing and this lesson often uses nano and cat in the same code block without the bit in the middle. It is definitely more time consuming to separate but I think also easier for beginners to follow.

### DIFF
--- a/episodes/04-changes.md
+++ b/episodes/04-changes.md
@@ -579,13 +579,7 @@ Two important facts you should know about directories in Git.
   in otherwise empty directories. The sole purpose of `.gitkeep` files is to populate a directory so that Git adds it to the repository. The name `.gitkeep` is just a convention, and in fact, you can name these files anything you like.
 
 2. If you create a directory in your Git repository and populate it with files,
-  you can add all files in the directory at once by using this command syntax:
-  
-  ```
-  git add <directory-with-files>
-  ```
-  
-  Try it for yourself:
+  you can add all the files in the directory at once by referring to the directory in your `git add` command. Try it for yourself:
   
   ```bash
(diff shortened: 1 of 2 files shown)


martinosorb:
I totally agree on the changes to 06-ignore, that's a clear improvement, thanks.

Regarding the change to 04-changes, why do you say that that's not bash? Is it because of the `<directory-with-files>` placeholder? If you think that's confusing, we could just change the whole thing to "you can add all files in the directory at once by using `git add`", because there's an example immediately after anyway. In fact, I think we should. What do you think?

erinmgraham:
Haha, I agree! I did consider it at the time but trying to be mindful of previous maintainer's decisions.

martinosorb:
Thanks!

PR #1064, opened 2025-01-21
include global flag for config list command

### LINKED ISSUE #907: add --global to listing the git settings
While teaching the git lesson (Setting Up Git) we noticed that the following command resulted in some confusion:

```
git config --list
```

For Windows users this command shows many more system settings than for Linux or Mac users, which made people wondering whether there was something wrong with their settings. Perhaps it is better to specify to only show the global settings:

```
git config --list --global
```

erinmgraham:
Closes #907

Using global flag to write config, we should use it read config.


martinosorb:
I know nothing about this, on my system it does the same, so I see no harm.

PR #1065, opened 2025-01-21
Update 01-basics.md

### LINKED ISSUE #941: Link to official Mercurial docs instead of our lesson

The [git lesson intro](https://swcarpentry.github.io/git-novice/01-basics.html#the-long-history-of-version-control-systems) refers to Mercurial as an alternative for version management and links to a very old SWC lesson (has not been updated in 6+ years).  Consider linking to the actual [Mercurial site](https://www.mercurial-scm.org/) instead (which is also old but...).

https://swcarpentry.github.io/git-novice/01-basics.html#the-long-history-of-version-control-systems

erinmgraham:
Closes #941 
Provide wiki link for git and mercurial and mention old carpentry lesson

### DIFF
--- a/episodes/01-basics.md
+++ b/episodes/01-basics.md
@@ -76,7 +76,7 @@ are *distributed*, meaning that they do not need a centralized server to host th
 These modern systems also include powerful merging tools that make it possible for multiple authors to work on
 the same files concurrently.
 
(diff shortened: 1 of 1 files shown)


martinosorb:
Just nitpicking, but as far as I recall there was a style recommendation to use "The Carpentries has" (singular and both capitalized, as the name of a single entity) rather than "the Carpentries have". I can't find it at the moment, but the website does that.

PR #1078, opened 2025-03-26
Fix touch command

lexnederbragt:
Unless I am mistaken, this change is needed for this command to work (and make sense).


martinosorb:
That makes sense.  Thank you.

PR #1081, opened 2025-03-28
Replace 'receipts' with 'pictures' to reduce issue for people with dyslexia

### LINKED ISSUE #1048: recipes, receipts may be problematic for people with dyslexia

I had a learner who has dyslexia and does not have English as their first language. They reported that the words 'recipes' and 'receipts' are problematic as they look very similar for them. If this is a widespread problem, perhaps the 'receipts' folder could get a name that is more dissimilar from 'recipes'?

lexnederbragt:
Closes #1048.

### DIFF
--- a/episodes/06-ignore.md
+++ b/episodes/06-ignore.md
@@ -145,7 +145,7 @@ Ignored files:
         a.png
         b.png
         c.png
-        receipts/
+        pictures/
 
 nothing to commit, working tree clean
 ```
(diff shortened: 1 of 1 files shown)


lexnederbragt:
@erinmgraham Thanks, I had totally not seen these. I have now replaced all occurrences and adjusted the challenges so that they hopefully still make sense.

erinmgraham (approved):
I can't help but notice that the data and plots dirs were renamed in challenge 1 to cake and pizza and then back to data and plots for a later challenge but that shouldn't have a major impact :-)

PR #1082, opened 2025-03-28
Improve language around ssh keys

lexnederbragt:
A learner for whom English is not their native language asked me what the word 'audit' means in this section. This PR attempts to avoid the word and rather explain what is meant and why.


erinmgraham (approved):
Looks good to me! Thanks @lexnederbragt

Issue #1022, opened 2024-10-14
explain branches


jorobledo:
### How could the content be improved?

Hi,
I read the content of the course, which I found very good! Nice job on that. But I found an omission (at least from what I could inspect) in explaining what branches are and how to use them. Is this on purpose? I think it is a very useful feature that git provides and that it should also be taught. I would be willing to develop some content for the course if there is interest. 

Please let me know you opinion!

martinosorb:
Hi, thanks for this. I am personally in favour of adding branches, however this is an issue that has been debated and is currently controversial. We can keep this issue open for further discussion.

jorobledo:
I am still very much of the opinion that git users should know branching. 

Would it make sense to do an initial PR to put the discussion forward and work on top of this initial version?

martinosorb:
@swcarpentry/git-novice-maintainers : I would be in favour of adding an 'extra' episode that could give an overview of branching, in a way that could be skipped if needed, so that the instructors can mix and match based on how much time they have and on what specific needs their audience has. Would that be possible?

erinmgraham:
I agree that adding an episode on branching would be very valuable. Given the interest and previous requests (issue #827, issue #918 [closed]), it makes sense to move forward with this. Toby noted there is a [lesson in the incubator](https://carpentries-incubator.github.io/git-novice-branch-pr/07-branches/) that might be a suitable starting point. And perhaps @davidwilby, @aragilar, and @marcodelapierre are still willing to help?

marcodelapierre:
At this stage I don't have bandwidth to contribute to this, apologies.

jorobledo:
Then we can probably do an extra material at the moment, where we explain what branching is, and then we can decide where to put it and what to trim from the current curricula so that it fits with the expected time (as suggested in those other issues). Also this would give instructors the possibility to refer to a carpentries material in case of questions or interest about branching from the audience (has happened to me).

For me, explaining collaborating through GitHub and not branching is just not right at the moment, because they will need it to contribute to others repositories.

davidwilby:
@erinmgraham Absolutely, I agree that the incubator lesson you link is a good starting point, I've taught this lesson a few times in order to include branching.

davidwilby:
I'm going to get started having a go at this and will open a PR if that's alright with you folks?

davidwilby:
I've opened a draft PR as a starting point for discussion: #1083

Just to reiterate the discussion points in case this is a better place to cover some of them:

[...]

PR #1083, opened 2025-04-16
Add episode on branching

davidwilby:
This draft PR is intended as a starting point for discussion on this topic as requested in #1022 

I've added the episode on branches from https://github.com/carpentries-incubator/git-novice-branch-pr/ more or less _verbatim_ albeit with name and format changes to match the new narrative. This is just added as `06b-branches.md` for now to avoid complicating the diff by changing all the other file numbers but showing where it would fit into the order.

Points of discussion:
+ Timing: I'm reluctant to contribute to the bloat of lessons, though I think this is a very important topic to include in this lesson.
+ Should pull requests also be covered in this lesson? They are already covered with a good example in the incubator lesson linked above.
+ How/should the originators of the material used from the incubator lesson be credited here?
+ This branching episode uses examples of a python and a bash version of an analysis as its branches - would it be preferable to use something more aligned with the food/recipe themes instead?

Looking forward to a lively discussion on this and thanks for the opportunity to contribute!

### DIFF
--- a/episodes/06b-branches.md
+++ b/episodes/06b-branches.md
@@ -1,0 +1,329 @@
+---
+title: Branches
+teaching: 20
+exercises: 0
+---
+
+::::::::::::::::::::::::::::::::::::::: objectives
+
+- Understand why branches are useful for:
+  - Working on separate tasks in the same repository concurrently
+  - Trying multiple solutions to a problem
+  - Check-pointing versions of code
+- Merge branches back into the main branch
+
+
+::::::::::::::::::::::::::::::::::::::::::::::::::
+
+:::::::::::::::::::::::::::::::::::::::: questions
+
+- What are branches?
+- How can I work in parallel using branches?
+
+::::::::::::::::::::::::::::::::::::::::::::::::::
+
+So far we've always been working in a straight timeline.
+However, there are times when we might want to keep
+our main work safe from experimental changes we are working on.
+To do this we can use branches to work on separate tasks in parallel
+without changing our current branch, `main`.
+
+We didn't see it before but the first branch made is called `main`.
+This is the default branch created when initializing a repository and
+is often considered to be the "clean" or "working" version of a
+repository's code.
+
+We can see what branches exist in a repository by typing
+
+```bash
+$ git branch
+```
+
+```output
+* main
+```
+
+
+The `*` indicates which branch we are currently on.
+
+In this lesson, Alfredo is trying to run an analysis
+and doesn't know if it will be faster in bash or python.
+To keep his main branch safe he will use separate branches
+for both bash and python analysis.
+Then he will merge the branch with the faster script
+into his main branch.
+
+First let's make the python branch.
+We use the same `git branch` command but now add the 
+name we want to give our new branch
+
+```bash
+$ git branch pythondev
+```
+
+We can now check our work with the `git branch` command.
+
+```bash
+$ git branch
+```
+
+```output
+* main
+  pythondev
+```
+
+We can see that we created the `pythondev` branch but we
+are still in the main branch.
+
+We can also see this in the output of the `git status` command.
+
+```bash
+$ git status
+```
+
+```output
+On branch main
+nothing to commit, working directory clean
+```
+
+To switch to our new branch we can use the `checkout` command
+we learned earlier and check our work with `git branch`.
+
+```bash
+$ git checkout pythondev
+$ git branch
+```
+
+```output
+  main
+* pythondev
+```
+
+Before we used the `checkout` command to checkout a file from a specific commit
+using commit hashes or `HEAD` and the filename (`git checkout HEAD <file>`). The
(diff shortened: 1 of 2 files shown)


episodes/06b-branches.md
```diff
@@ -0,0 +1,329 @@
...
+* main
+```
+
+
+The `*` indicates which branch we are currently on.
+
+In this lesson, Alfredo is trying to run an analysis
+and doesn't know if it will be faster in bash or python.
+To keep his main branch safe he will use separate branches
+for both bash and python analysis.
```
martinosorb, line 52: I think it would be good to keep the "flavour" of Alfredo as a chef keeping his recipes under version control. So perhaps we could say he's experimenting with a new version of his famous strawberry cream cake in which he uses mango and less sugar, or something of the kind?

episodes/06b-branches.md
```diff
@@ -0,0 +1,329 @@
...
+$ git status
+```
+
+```output
+On branch main
+nothing to commit, working directory clean
+```
+
+To switch to our new branch we can use the `checkout` command
+we learned earlier and check our work with `git branch`.
```
martinosorb, line 90: This has been changed, and we now try not to use `checkout`. The command we can use to move to another branch is `git switch`. We should teach that, even if it's newer, because `checkout` does many things and is confusing to beginners.

episodes/06b-branches.md
```diff
@@ -0,0 +1,329 @@
...
+
+```output
+  main
+* pythondev
+```
+
+Before we used the `checkout` command to checkout a file from a specific commit
+using commit hashes or `HEAD` and the filename (`git checkout HEAD <file>`). The
+`checkout` command can also be used to checkout an entire previous version of the
+repository, updating all files in the repository to match the state of a desired commit.
```
martinosorb, line 105: Same as above: `checkout` is no longer taught in this lesson, so this is not relevant any more. The comparison can be done with `git restore` if needed.

episodes/06b-branches.md
```diff
@@ -0,0 +1,329 @@
...
+some changes are committed to our new branch.
+
+Now lets make our python script.  
+
+For simplicity, we will `touch` the script making an empty file
+but imagine we spent hours working on this python script for our analysis.
+
+```bash
+$ touch analysis.py
+```
```
martinosorb, line 124: I would make an experimental edit to an existing recipe instead.

episodes/06b-branches.md
```diff
@@ -0,0 +1,329 @@
...
+Since we've never merged the changes from the `bashdev` branch,
+git warns us about deleting them and tells us to use the `-D` flag instead.
+
+Since we really want to delete this branch we will go ahead and do so.
+
+```bash
+git branch -D bashdev
+```
+
+```output
```
martinosorb, line 327: It's missing the ending (usually each episode has a summary at the end)

martinosorb:
Thank you so much @davidwilby!
I'd love to see something similar.

I've left comments inline above. Summarizing and adding more general thoughts:
- Flavour: I would keep something more in line with the current theme of recipes. An experimental version of an existing recipe could work,
- `git checkout`: no longer taught. Let's use `git switch` instead, and compare with `git restore`.
- The concluding summary is missing
- Rather than having _two_ new branches, each with a new file, and then merging one of them, I would suggest we only use _one_ branch, do all the diffs, switch back and forth, and then decide to merge. Don't you think? I can also think of advantages of the current setup, to be honest, so let's discuss.
- We should have a note to the instructors saying that this can be skipped if not relevant.

episodes/06b-branches.md
```diff
@@ -0,0 +1,329 @@
+---
+title: Branches
+teaching: 20
```
jorobledo, line 3: I'm quite sure this can take 25 to 30 minutes. Also, I believe it important to give them an exercise to try it out themselves, because it's something that can be confusing and just trying it allows to fully understand what one is doing. maybe 5 to 10 for exercises?

episodes/06b-branches.md
```diff
@@ -0,0 +1,329 @@
...
+
+- What are branches?
+- How can I work in parallel using branches?
+
+::::::::::::::::::::::::::::::::::::::::::::::::::
+
+So far we've always been working in a straight timeline.
+However, there are times when we might want to keep
+our main work safe from experimental changes we are working on.
+To do this we can use branches to work on separate tasks in parallel
```
jorobledo, line 28: we can branch our timeline to work on separate tasks in parallel

jorobledo (commented):
added inline comments and answers to comments to attempt a more "in flavour" approach. Also, I propose to add a merge conflict example and solution, because it is something they will need to do quite early on their git experience with branches.

episodes/06b-branches.md
martinosorb: That's pretty much what I would suggest. But I think we need a good discussion on whether to do one branch or two branches, I'm not certain of my opinion.

PR #1088, opened 2025-05-27
Explain how to use config for custom key names and paths

### LINKED ISSUE #846: Mention the importance of using the default SSH key location
In episode 7, we state: "We want to use the default file, so just press Enter."

At a recent Carpentries UK community call, one instructor mentioned that they derailed a lesson because they chose a non-default location for the key, because of the specific setup of the computer they were using. However, OpenSSH only looks for keys in the default locations.

If this is something that other instructors have encountered difficulty with, is it worth adding a sentence after the one quoted above? Something along the lines of "(If you choose another location here, then the key won't automatically be detected when you come to use it. However, you can use the `-i` option to specify it explicitly.)"

erinmgraham:
Closes #846

I encountered the issue teaching on Mon and needed to generate custom named keys. I added a spoiler with the steps to add git as host to a config file because I think that will be the easiest option.

### DIFF
--- a/episodes/07-github.md
+++ b/episodes/07-github.md
@@ -168,6 +168,36 @@ ls: cannot access '/c/Users/Alfredo/.ssh': No such file or directory
 If SSH has been set up on the computer you're using, the public and private key pairs will be listed. The file names are either `id_ed25519`/`id_ed25519.pub` or `id_rsa`/`id_rsa.pub` depending on how the key pairs were set up.
 Since they don't exist on Alfredo's computer, he uses this command to create them.
 
+:::::::::::::::::::::::::::::::::::::::::  spoiler
+
+## Resolving SSH Key Conflicts: Custom Names and Paths
+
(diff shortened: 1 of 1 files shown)


martinosorb:
Hi Erin. Do you  think this is a common problem? Because the answer is rather technical and may confuse people with things they have no idea about, or could give the impression that this is complicated stuff. Maybe it's better to add this to the instructor notes instead, as @kekoziar suggests in the related issue? Thanks for taking care of this

erinmgraham:
Thanks for the discussion Martin! While it may not be a common issue, it can be a blocker—and instructor notes are not easily accessible in the moment. Having it as spoiler keeps the main flow clean while offering help right where it’s needed, without overwhelming beginners. The content may be technical but I wouldn't say it's confusing. 

That said, I’m happy to get a third or fourth opinion if others (@kekoziar @edbennett) feel strongly—just want to make sure learners aren’t stuck without a clear path forward.

PR #1098, opened 2025-10-29
DOC: Fixed text mismatch in markdown section

Prabhu-LANL:
Fixing a small error in the text of one of the tutorials.


erinmgraham (approved):
Good catch, @Prabhu-LANL! Thank you for the taking the time to submit a fix :-)

Issue #1020, opened 2024-10-08
Explain the former use of checkout


martinosorb:
### How could the content be improved?

Add a comment in the instructor notes explaining that we used to use 'checkout' instead of 'switch' and 'revert', and this should still be used in older versions of git.

### Which part of the content does your suggestion apply to?

05 - Exploring history

PR #1109, opened 2026-01-29
Add note on 'git restore' to Instructor Notes

gray95:
This PR addresses issue #1020 by adding a note on `git restore` to `instructor-notes.md` clarifying its relation to `git checkout`.

Info on these changes can be found [here](https://www.infoq.com/news/2019/08/git-2-23-switch-restore), [here](https://stackoverflow.com/questions/61130412/what-is-the-difference-between-git-checkout-vs-git-restore-for-reverting-un), and in Git's [release notes](https://github.com/git/git/blob/master/Documentation/RelNotes/2.23.0.adoc).

`git switch` is not mentioned in the lesson to the best of my knowledge so I didn't mention it even though it was introduced at the same time as `git restore`.

### DIFF
--- a/instructors/instructor-notes.md
+++ b/instructors/instructor-notes.md
@@ -158,6 +158,10 @@ working in teams or not, because it is
 - This is a good moment to show a log within a Git GUI. If you skip it
   because you're short on time, show it once in GitHub.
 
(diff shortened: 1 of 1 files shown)


erinmgraham:
Looks good to me too!

PR #1112, opened 2026-03-04
scale image sizes

martin-raden:
The images to illustrate the need for version control were not scaled.

I added appropriate width values.

Best Martin

### DIFF
--- a/episodes/01-basics.md
+++ b/episodes/01-basics.md
@@ -38,20 +38,20 @@ think of it as a recording of your progress: you can rewind to start at the base
 document and play back each change you made, eventually arriving at your
 more recent version.
 
-![](fig/play-changes.svg){alt='A diagram demonstrating how a single document grows as the result of sequential changes'}
+![](fig/play-changes.svg){alt='A diagram demonstrating how a single document grows as the result of sequential changes' width="40%"}
 
 Once you think of changes as separate from the document itself, you
 can then think about "playing back" different sets of changes on the base document, ultimately
 resulting in different versions of that document. For example, two users can make independent
 sets of changes on the same document.
 
-![](fig/versions.svg){alt='A diagram with one source document that has been modified in two different ways to produce two different versions of the document'}
+![](fig/versions.svg){alt='A diagram with one source document that has been modified in two different ways to produce two different versions of the document' width="60%"}
 
 Unless multiple users make changes to the same section of the document - a 
 [conflict](../learners/reference.md#conflict) - you can
 incorporate two sets of changes into the same base document.
 
-![](fig/merge.svg){alt='A diagram that shows the merging of two different document versions into one document that contains all of the changes from both versions'}
+![](fig/merge.svg){alt='A diagram that shows the merging of two different document versions into one document that contains all of the changes from both versions' width="60%"}
 
 A version control system is a tool that keeps track of these changes for us,
 effectively creating different versions of our files. It allows us to decide


erinmgraham:
Thanks for working on improving the presentation of these diagrams — the progression you’re trying to show makes sense. 

I built a local copy and noticed that the first image in the sequence (play-changes.svg) becomes visually distracting with this change. Because the SVGs have very different aspect ratios, scaling them to be roughly similar in width affects the first image disproportionately (screenshot below).

We generally avoid explicit image sizing in lessons and let the CSS handle scaling by default to avoid issues like these and ease maintainability.

I'll wait to see what others think but if scaling these diagrams is preferred, I think a more robust approach would be to adjust the SVGs themselves (e.g. normalising the canvas/viewBox so they have similar proportions). That keeps the lesson markup simple and avoids layout tweaks that fight the theme. Happy for you to go that route if you’d like.

screenshot_git_swc_resizePR

martin-raden:
Hi,

I am surprised about the rendering differences. I have used this part of your lesson within my own carpetries-based tutorial (with proper citation) https://dr-eberle-zentrum.github.io/intro-to-git-and-github/instructor/01-git-basics.html and it scales just fine:

grafik

Personally, I find the images in the original sizes much too big given their low (but important!) information content.

I see your point in the non-scaling strategy, but eventually I would says this is exactly what stylesheets etc are for?!!

Will see if I find the time to alter the svgs instead.

best,
Martin

martin-raden:
as an alternative you can put a fixed `width` into the svgs at line 5 (within the head `<svg>` definition after the `viewport`):

- `versions.svg` and `merge.svg` : line 5 change at the end to `viewBox="0 0 384 354" width="300px"`
- `play-changes.svg` : line 5 change at the end to `viewBox="0 0 615 196.5" width="470px"`

and the files will be scaled to similar dimensions.

Hope that helps,
Martin

erinmgraham:
Hi Martin, 

Thank you for the detailed follow-up — I appreciate you thinking through alternatives.

This is not a discussion about whether the images are too large, it's a discussion about where responsibility for layout lives (ie where the fix should occur).

I am not comfortable with adding fixed widths (either in Markdown or inside the SVGs), as that couples these assets to a particular rendering context which a lesson cannot control. Case in point is the difference in our respective screenshots.

I would be happy if the SVG canvases/viewBoxes were adjusted so they scale consistently under the existing CSS and I think that would be a solid improvement. Otherwise, I’m inclined to keep the current behaviour.

martin-raden:
Hi,

sure, your repo.

Eventually, I dont see a difference between a absolute change of the viewport (ie the svg-internal coordinate system) and setting the intended svg-width that does the standard scaling of the viewport to the targeted width. Since SVG is a vector graphics, such a scaling is done anyway and intended.

So from my perspective, setting width equals to scaling a pixelgraphics to the respective dimension but without changing the internal coordinate system, which might cause rounding issues and thus results in a distorted image..

Eventually, I went for the svg-width solution in my repo (without the markdown-css-scaling suggested first).

I can rewrite this PR towards this as well if you like. Otherwise I am not offended when this PR is closed without merging. 😉 
It's just a suggestion since I, personally, think the images way to big right now. But it's your decision in the end.

Best,
Martin

erinmgraham:
Asked Carpentries' Workbench team for best practice advice on this and Toby was the only one who piped up with "have certainly used the same approach as in that PR". Merging.

PR #1114, opened 2026-03-18
keep comic light mode

martin-raden:
Hi @froggleston 

thanks for the comments in PR #1113 

given that, find attached the duplication of the comic to keep it light in dark mode.

personally, I prefer also the merge, conflict, .. SVGs suited for keeping them light, but since the [thread you linked](https://github.com/carpentries/varnish/pull/153) leaves them dark, I left them out for now.

If the maintainers are interested, I can add `-dark` duplicates for these files too.

Best,
Martin


froggleston:
Thanks @martin-raden ! I'll let the lesson maintainers decide from here :)

erinmgraham:
Thanks @martin-raden — I’m happy to merge this. Keeping the comic light in dark mode makes sense, as it would be difficult to recreate faithfully with a dark palette.

For the other images I’m conscious that dark‑mode users aren’t all the same: some are fine with light figures, but others do prefer content that’s explicitly designed for dark backgrounds. These diagrams are simpler and  arguably better candidates for explicit dark‑mode color conversions.

That said, I'm not a dark mode user and welcome all contributions :-)

Issue #1080, opened 2025-03-27
clarify git diff behavior when HEAD is omitted


erinmgraham:
> Maybe related, I stumbled upon the following sentence today:
> 
> Note that `HEAD` is the default option for `git diff`, so omitting it will not change the command’s output at all (give it a try).
> 
> I found this to be very confusing, because the output is only the same if no additional files have been staged since the last commit. 

 _Originally posted by @fkohrt in [#769](https://github.com/swcarpentry/git-novice/issues/769#issuecomment-2745871257)

erinmgraham:
Hi @fkohrt, thank you for your contribution! I will make a new issue with your comment because the text you referenced is found in episode 05-history and issue 769 relates to episode 04-changes. 

Are you able to suggest alternative text that would be less confusing?

fkohrt:
Maybe as follows?

> If run without arguments, `git diff` compares the files in your working directory with what is about to be save permanently (the staging area). `git diff HEAD` compares the files in your working directory with the most recently committed version instead. The real power of `git diff` lies in its ability to compare with previous commits.

Displaying this comparison in another diagram like the one by @astroDimitrios [here](https://github.com/swcarpentry/git-novice/issues/769#issuecomment-2271762658) might additionaly be helpful.

PR #1116, opened 2026-04-08
Update git diff explanation in history.md

johnschoneboom:
Clarified the explanation of the default option for 'git diff' and its behavior with staged files.

_If this pull request addresses an open issue on the repository, please add 'Closes #NN' below, where NN is the issue number._

_Please briefly summarise the changes made in the pull request, and the reason(s) for making these changes._

The Exploring History lesson asserted that HEAD was the default option for 'git diff', which is only true if the staging area is empty. If there are staged changes, 'git diff' compares the working area to the staged area, and adding HEAD will give a different result.

_If any relevant discussions have taken place elsewhere, please provide links to these._

<details>

For more guidance on how to contribute changes to a Carpentries project, please review [the Contributing Guide](CONTRIBUTING.md) and [Code of Conduct](https://docs.carpentries.org/topic_folders/policies/code-of-conduct.html).

Please keep in mind that lesson Maintainers are volunteers and it may be some time before they can respond to your contribution. Although not all contributions can be incorporated into the lesson materials, we appreciate your time and effort to improve the curriculum. If you have any questions about the lesson maintenance process or would like to volunteer your time as a contribution reviewer, please contact The Carpentries Team at team@carpentries.org.

</details>

### DIFF
--- a/episodes/05-history.md
+++ b/episodes/05-history.md
@@ -62,7 +62,7 @@ index b36abfd..0848c8d 100644
 +An ill-considered change
 ```
 
-Note that `HEAD` is the default option for `git diff`, so omitting it will not change the command's output at all (give it a try). However, the real power of `git diff` lies in its ability to compare with previous commits. For example, by adding `~1` (where "~" is "tilde", pronounced [**til**\-d*uh*]), we can look at the commit before `HEAD`.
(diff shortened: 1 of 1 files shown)


erinmgraham:
Thank you for suggesting alternative text for the git diff explanation, I appreciate the thought you put into this.

After digging into this further, including revisiting issue #1080, I think the better approach is to simplify the text here and clarify git diff more explicitly in episode 04, where it’s introduced. This episode is intended to focus on history and the use of identifiers and trying to capture the full variability of the git diff command here muddles the message. 

I’m going to close this PR and follow up with a new one that links back here so your contribution and the discussion aren’t lost :-)

Issue #1121, opened 2026-04-29
Clarify email address requirement for ssh-keygen -C in Episode 7


gjduine:
## Problem
The example in [Episode 7, line 206](https://github.com/swcarpentry/git-novice/blob/3e9bb3c8b01f5ca6e6ac96e06d252116f8db520c/episodes/07-github.md?plain=1#L206) shows:

```bash
$ ssh-keygen -t ed25519 -C "a.linguini@ratatouille.fr"

For beginners, it's unclear whether the email must be their GitHub account email or if any email works.

Suggested Improvement
Add a note clarifying that the -C flag adds a comment/label to identify the key:

Note: The email address in the -C option is used as a label to identify this SSH key. Use the email address associated with your GitHub account so you can easily remember which key belongs to which account.

This will help new learners choose the correct email without confusion.

### Which part of the content does your suggestion apply to?

episode 7 in git-novice

erinmgraham:
Hi @gjduine, thank you for raising this issue - it's good to be thinking about beginner clarity.

You’re absolutely right that the -C flag is used to add a comment/label to the SSH key. One important nuance, though, is that there isn’t a required or correct email to use there. The official documentation doesn't prescribe what to use here and while many people, including us, use their github email address because it's convenient, others may use hostnames, the key's purpose, or no comment at all. 

Taking that into consideration, can I ask you to revise your suggested improvement text and submit it as a PR for review?

gjduine:
OK, I just submitted a PR

PR #1122, opened 2026-05-01
clarification on -C option

gjduine:
Closes #1129

For novice users, the addition of an e-mail address is unclear, so some additional information would be helpful

Full issue here:
https://github.com/swcarpentry/git-novice/issues/1121

### DIFF
--- a/episodes/07-github.md
+++ b/episodes/07-github.md
@@ -205,6 +205,10 @@ To create an SSH key pair Alfredo uses this command, where the `-t` option speci
 ```bash
 $ ssh-keygen -t ed25519 -C "a.linguini@ratatouille.fr"
 ```
+Notes: 
+1) The -C flag adds a comment/label to the SSH key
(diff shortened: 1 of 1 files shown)


martinosorb:
Thank you for this. It looks good in principle, but I find it a little dry: for example it's not clear to me what "hostnames, or no comment at all" means. Maybe we can make it into a friendlier sentence? Something like
"In the command above, the `-C- flag is used to label the SSH key with the owner's email address. It's good practice to use the same email associated to your GitHub account, but it does not have to be, and will not be used to validate the access. You can also decide not to add any label."

How does that sound?

gjduine:
yes that sounds much better, in fact. Thank you. How does it work from here? Do I change the sentence?

martinosorb:
You can add another commit to the branch where you did the first change, and it will automatically be updated in the pull request. Thanks!

PR #1125, opened 2026-06-17
Modify the solution to Conflicts on Non-textual Files challenge so that it is more consistent with the intended behavior

Zhuminem:
When trying to keep both images in this challenge, the original solution suggests using "git mv" for renaming one conflicted image file, but only "mv" for another.

Using "mv" in both cases will be more consistent and still correct, while preventing git from auto-resolving the conflict and barring a "git checkout" navigation to the alternate path -- it happens sometimes.

The original solution should still work in some cases, but the updated solution should work in all cases.

### DIFF
--- a/episodes/09-conflict.md
+++ b/episodes/09-conflict.md
@@ -504,7 +504,7 @@ image and rename it:
 
 ```bash
 $ git checkout HEAD guacamole.jpg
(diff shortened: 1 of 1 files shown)


erinmgraham:
Great catch! Thank you for your contribution :-)

